PR #1, opened 2017-04-03
First implementation

delvedor:
As titled, this is the very first implementation of the swagger plugin.
As usual, I've created a PR instead of push directly the code to master to discuss which is the best implementation and because the code is far from finished.

cc @mcollina 

TODO: 
- [x] basic implementation
- [x] serve swagger data in a route
- [ ] more test
- [x] *automatic generation of swagger UI (?)*
- [x] *serve the UI as a route (?)*


delvedor:
@mcollina generate the UI is not a trivial thing, I think we can merge this and then think about how support the UI generation as well.

delvedor:
Now we are exposing the `/documentation` route and returning by default a json object with the swagger declaration.

Probably this not cover the entire [swagger specification](https://swagger.io/specification/), but I think is a good starting point.

README.md
```diff
@@ -1,2 +1,98 @@
...
+  swagger: {
+    info: {
+      title: 'Test swagger',
+      description: 'testing the fastify swagger api',
+      version: '0.1.0'
+    },
+    host: 'localhost',
+    schemes: ['http'],
+    consumes: ['application/json'],
+    produces: ['application/json']
```
mcollina: are these the default?

.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: node_js
+node_js:
+  - "7"
```
mcollina: node 8 is missing

index.js
```diff
@@ -0,0 +1,224 @@
...
+    for (var node of fastify) {
+      // The node path name
+      const url = formatParamUrl(Object.keys(node)[0])
+      // object with all the methods of the node
+      const routes = node[Object.keys(node)[0]]
+      const swaggerRoute = {}
+      swaggerRoute[url] = {}
+
+      // let's iterate over the methods
+      Object.keys(routes).forEach(method => {
```
mcollina: I think you might have way better perf using a `for`  loop. You are allocating a new function for each node in fastify.

index.js
```diff
@@ -0,0 +1,224 @@
...
+        const route = routes[method]
+        swaggerRoute[url][method] = {}
+
+        const parameters = []
+        const schema = route.schema.schema
+
+        // All the data the user can give us, is via the schema object
+        if (schema) {
```
mcollina: A nice feature would be to hide a route, even if it has a schema.

index.js
```diff
@@ -0,0 +1,224 @@
...
+          properties: {
+            yaml: { type: 'boolean' }
+          }
+        },
+        response: {
+          200: {
+            description: 'The swagger definition of the routes',
+            type: 'object',
+            properties: {
+              hello: { type: 'string' }
```
mcollina: this is probably no true.

index.js
```diff
@@ -0,0 +1,224 @@
...
+            }
+          }
+        }
+      }
+    }, function (req, reply) {
+      if (req.query.yaml) {
+        return reply
+          .header('Content-Type', 'text/plain')
```
mcollina: this is probably not the mime type

index.js
```diff
@@ -0,0 +1,224 @@
...
+          .send(fastify.swagger({ yaml: true }))
+      }
+      reply.send(fastify.swagger())
+    })
```
mcollina: we should probably have some sort of caching in place.

README.md
delvedor: No, is just an example for the general configuration :)

index.js
delvedor: Ooooops :D

README.md
```diff
@@ -1,2 +1,98 @@
...
+
+## Install
+```
+npm i fastify-swagger --save
+```
+
+## Usage
+Add it to your project with `register` and pass it some basic options, then call the `swagger` api and you are done!
+
+*Note that currently only the generation of yaml/json files is supported, the swagger UI will not be generated.*
```
mcollina: do you plan to add the swagger-ui here?

README.md
mcollina: Then those would need to be documented somewhere, as they are not clear.

index.js
delvedor: https://www.quora.com/What-is-the-correct-MIME-type-for-YAML-documents
Thoughts?

index.js
delvedor: We can use a custom property,`hide: true`, for example?

README.md
delvedor: Yeah, I must update the whole docs.

index.js
mcollina: I would go for `application/x-yaml` with an explanation and a link to the IANA registry.

README.md
delvedor: I'd like to, but I must figure out how to do it.

index.js
```diff
@@ -30,25 +30,32 @@ function fastifySwagger (fastify, opts, next) {
         },
         response: {
           200: {
-            description: 'The swagger definition of the routes',
-            type: 'object',
-            properties: {
-              hello: { type: 'string' }
-            }
+            description: 'The swagger definition of the routes'
```
mcollina: I would hide this by default.

delvedor: We can't, the response **must** have the description property, if we don't set it by hand, this plugin will set it at `"Default response"`.

mcollina: I mean with `hide: true`, so that it does not show up.

README.md
allevo: Is it done?

README.md
```diff
@@ -1,2 +1,107 @@
...
+      properties: {
+        hello: { type: 'string' }
+      }
+    }
+  }
+}, (req, reply) => {})
+
+fastify.ready(err => {
+  if (err) throw err
+  fastify.swagger()
```
allevo: Why this?
Can this plugin attach to `ready` avvio event to generate the swagger?

I don't remember if `fastify` allow it, but what happen if a route is attached after the start up? Can be rebuild with a `{ force: true }`?

delvedor: `fastify.swagger()` is a synchronous api, we can't use `.ready` inside as we did for `inject`, otherwise send the result inside a route will be painful.

> I don't remember if fastify allow it, but what happen if a route is attached after the start up? Can be rebuild with a { force: true }?

I didn't try, but I don't think it's a good idea. Furthermore, which is the use case?

README.md
allevo: @delvedor here!

README.md
allevo: > I didn't try, but I don't think it's a good idea. Furthermore, which is the use case?

no one. Please update the doc

delvedor:
I've added a basic support for the swagger UI generation.
Please review. Every suggestion is well accepted :)

routes.js
```diff
@@ -0,0 +1,75 @@
+'use strict'
+
+const fp = require('fastify-plugin')
+const readFileSync = require('fs').readFileSync
+const resolve = require('path').resolve
+
+const files = {
+  html: readFileSync(resolve('static', 'index.html'), 'utf8'),
```
allevo: Truly this may wrong. `resolve` resolves the path from your `cwd` (the application path)

Maybe you can use this https://github.com/fastify/fastify-static

.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: node_js
+node_js:
+  - "8"
```
mcollina: can you add node 9?

routes.js
```diff
@@ -0,0 +1,75 @@
...
+      case 'swagger-ui-bundle.js':
+        return reply.type('application/javascript').send(files.js)
+
+      case 'swagger-ui-standalone-preset.js':
+        return reply.type('application/javascript').send(files.preset)
+
+      default:
+        return reply.code(404).send(new Error('Not found'))
+    }
+  }
```
mcollina: can you use fastify-static for this?

index.js
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/* function loadSwaggerUI (fastify, opts, next) {
+  swaggerTools.initializeMiddleware(opts.swaggerDoc, onMiddleware)
+
+  function onMiddleware (middleware) {
+    fastify.use(middleware.swaggerUi())
+    next()
+  }
+} */
```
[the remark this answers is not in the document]
delvedor: Ops! :D

routes.js
delvedor: I don't know if it is worth add a middleware just for serve this files.
Since this is just for the documentation I think this could be left as is.

PR #2, opened 2017-09-12
Some fixes

allevo:
Update fastify version.
Fix "undefined" support bug ([here](https://github.com/fastify/fastify-swagger/blob/first-implementation/index.js#L32) miss the `type` property)
Avoid to change schema of `Store` configuration

### DIFF
--- a/index.js
+++ b/index.js
@@ -242,4 +242,4 @@ function formatParamUrl (url) {
   }
 }
 
-module.exports = fp(fastifySwagger, '>=0.14.0')
+module.exports = fp(fastifySwagger, '>=0.27.0')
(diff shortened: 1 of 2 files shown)


allevo:
Uhm... I don't know: the previous code with the `fastify@0.27` doesn't work.
If you want I'll split this PR into 2 commit to see the break

mcollina:
Ok. Then let's increase the minimum version of fastify in what you pass in fastify-plugin.

allevo:
Please don't merge, example is broken.

Maybe some tests are needed

delvedor:
I've updated #1, thanks!

Issue #4, opened 2017-11-10
Hmm... Find a hard code here


qusiba:
https://github.com/fastify/fastify-swagger/blob/7d40a7b518c93e2fd43fd2f29a915a7c0dd2f88b/static/index.html#L75

qusiba:
Maybe can use this package https://www.npmjs.com/package/swagger-ui-dist

delvedor:
Pr are welcome :)

zenflow:
I think this can be closed. This URL is no longer hard-coded. This was addressed by PRs #7 & #9.

https://github.com/fastify/fastify-swagger/blob/b6b2b5fa3562cc7d75c5b06ca439d6cadfdb6fb9/static/index.html#L75

mcollina:
Fantastic!

PR #6, opened 2017-11-21
Adds support for OAS 3.0.0

nileshmali:
Adds support for OAS 3.0.0.

Fixes: #4


mcollina:
There are a lot of changes that are only about code formatting, can you please restore them as they were before?

nileshmali:
Sure. I will update.

zenflow:
@nileshmali Not sure what I'm missing here... What does OpenAPI Specification 3 have to do with issue #4 titled "Hmm... Find a hard code here"?

(Btw I think #4 can be closed, see comment there)

nileshmali:
@zenflow When I opened PR #4 was open. This PR adds support for OAS 3 and also fixes #4 plus this uses swagger-ui-dist to display swagger-ui.

nileshmali:
I will submit separate pull requests for each OAS 3 support and use of swagger-ui-dist.

zenflow:
@nileshmali Ahh I see.. the PR aimed at adding OAS 3 support and just incidentally you fixed #4.. cool

I think I see the value of adding OAS 3 support, but what is the value of using swagger-ui-dist? (May be good to put your answer in an issue)

PR #9, opened 2017-11-27
Support serving `/documentation` from prefixed path

zenflow:
Currently, if we try to register fastify-swagger inside a [route-prefixed plugin](https://github.com/fastify/fastify/blob/master/docs/Routes.md#route-prefixing), 
when we try to use the Swagger UI, we see in big letters "Failed to load spec." (and a message like `GET http://localhost:3002/documentation/json 404 (Not Found)` in the browser console).

This PR fixes the URL used to load the spec (`./documentation/json`), so that it's directory is the same as the other assets (e.g. `./documentation/swagger-ui.css`) which already work with prefixed paths.

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -72,7 +72,7 @@ window.onload = function() {
   
   // Build a system
   const ui = SwaggerUIBundle({
-    url: `${window.location.origin}/documentation/json`,
+    url: resolveUrl('./documentation/json'),
     dom_id: '#swagger-ui',
     deepLinking: true,
     presets: [
@@ -86,6 +86,12 @@ window.onload = function() {
   })
 
   window.ui = ui
+
+  function resolveUrl (url) {
+      const anchor = document.createElement('a')
+      anchor.href = url
+      return anchor.href
+  }
 }
 </script>
 </body>


delvedor:
This happens because if a plugin (such as `fastify-swagger`) is registered with `fastify-plugin`, the prefix option will not work.
You can try to implement a similar solution to the one we have adopted in [`fastify-bankai`](https://github.com/fastify/fastify-bankai) :)

zenflow:
@delvedor The problem has nothing to do with the fact that the `prefix` option doesn't work on plugins that use fastify-plugin

I am trying to use fastify-swagger *inside* a route-prefixed (non-`fastify-plugin`) plugin, e.g.

```
fastify.register(apiPlugin, {prefix: '/api'})
function apiPlugin (fastify, options, next) {
  fastify.route(someRouteDefinition)
  fastify.register(fastifySwagger, {exposeRoute: true}) // <---
  next()
}
```

This causes the swagger spec and all the assets for swagger-ui to be served from `/api/documentation` (using my `/api` prefix).

The only problem is the bit of code I fixed in `index.html` that is looking for the json swagger spec at the wrong URL. It does not account for the possibility that everything is being served from a prefixed URL.

This PR fixes that bit of code, to find the swagger spec at `./documentation/json` (relative path, note the `.`) instead of `/documentation/json` (absolute path). Of course in both cases we are actually using a fully-qualified URL with the origin included.

I've tested this fix with my project, and it allows me to serve the swagger-ui from `/api/documentation`.

PR #18, opened 2018-01-05
Security

scriptjs:
Fixes security definitions and scope to enable apiKeys, methods, and scopes.


scriptjs:
So this give us global security definition providing the authenticate button as below:

screen shot 2018-01-05 at 7 46 44 pm

Pressing it shows the security options as below:

screen shot 2018-01-05 at 7 47 21 pm

Route security give us links on the locks for each route enabling the override for the method as in below:

screen shot 2018-01-05 at 7 47 36 pm

scriptjs:
@delvedor @mcollina On the README, the big thing to remember is that this is documentation and you must still implement the security to mate with it. I can definitely see that something more tightly integrated with fastify-auth should be possible so you get both out of the box.

scriptjs:
Can we get this merged or is there anything else needed for this before we do? I'm looking to work out the issue with the responses next, then will prepare the module under the fastify-openapi name last.

mcollina:
Can you update the README in that direction?

scriptjs:
@mcollina Hi. Yup, just doing that now. will send another commit in a few minutes.

Issue #13, opened 2017-12-26
Declare an empty response for 204


tnguyen14:
For 204 response (empty response), Swagger allows for its description by not specifying a `content` https://swagger.io/docs/specification/describing-responses/.

However, for `fastify-swagger`, I haven't found a way to do that yet. If I do:

```js
{
  schema: {
    response: {
      204: {
        description: 'Successfully deleted item'
      }
    }
  }
}
```

I get the following stacktrace:

```
Uncaught Exception Error: undefined unsupported
    at build (/Users/tri/project/node_modules/fast-json-stringify/index.js:80:13)
    at getValidatorForStatusCodeSchema (/Users/tri/project/node_modules/fastify/lib/validation.js:12:10)
    at /Users/tri/project/node_modules/fastify/lib/validation.js:18:21
    at Array.reduce (<anonymous>)
    at getResponseSchema (/Users/tri/project/node_modules/fastify/lib/validation.js:17:22)
    at build (/Users/tri/project/node_modules/fastify/lib/validation.js:44:28)
```

mcollina:
This is likely a bug in fastify itself rather than this module. Would you like to submit it there?

tnguyen14:
Sure thing. I opened https://github.com/fastify/fastify/issues/609

cortopy:
@mcollina @tnguyen14 I'm also interested in getting this working for 302 responses. What does remain to be done? I can see this has been open for a while. Happy to complete or submit a new PR

mcollina:
I'm not sure, check https://github.com/fastify/fastify-swagger/pull/19.

If you want to make this moving again, feel free to send a new PR.

PR #19, opened 2018-01-07
fixes empty response #13

scriptjs:
This is a simple fix. The trouble stems from making response objects for swagger from fastify responses. Fastify responses require a type because they all handed off for validation. Having an undefined type is no dice for the validator but 'null' is valid.

Safe to say, empty responses don't have a type, so setting type to 'null' makes this explicit for fastify. ie.

    response: {
      204: {
        description: 'Successful deleted item',
        type: 'null'
      }
    },

When null is set for empty responses, this change ensures it is looked up and the type property deleted when converted to a swagger response. This leaves the description only for documentation that works for swagger.

screen shot 2018-01-06 at 11 26 53 pm

### DIFF
--- a/index.js
+++ b/index.js
@@ -213,6 +213,8 @@ function genResponse (response) {
 
   Object.keys(response).forEach(key => {
     if (response[key].type) {
+      if (response[key].type === 'null') delete response[key].type
+
       var rsp = response[key]
       var description = response[key].description
       var headers = response[key].headers


index.js
```diff
@@ -213,6 +213,8 @@ function genResponse (response) {
 
   Object.keys(response).forEach(key => {
     if (response[key].type) {
+      if (response[key].type === 'null') delete response[key].type
```
mcollina: can you duplicate the object if you plan to make edits?

scriptjs: Not sure what your mean. If you edit your code to provide an object, you get the result of the object in the swagger documentation which is also valid for swagger.

scriptjs: I can add a unit test, but we because we are dealing with the swagger UI, we need nightmare to verify results we are getting in the documentation. I can go ahead and do this if you wish.

scriptjs: If you are good with the change, I'll commit a piece to the README concerning empty responses as well.

mcollina: As this caused a crash when the route was accessed, just add a test that does not crash the server/throw exception when rendering.

As for creating a copy: just create a new jnstance with the property you need if you want to delete things.

If you need only a one-level copy you can use Object.assign({}, obj) (check this), or if you need a deep copy use the clone npm module.

scriptjs: kk

mcollina:
Any updates on this one? Is this still something you want to pursue?

avasuro:
Same issue - looks like in current version of fastify-swagger there is no ability to set type: 'null' or omit type property.

PR #21, opened 2018-01-14
remove dynamic requirement fixes #20

### LINKED ISSUE #20: Remove dynamic requirements
I use webpack on the server side. ES 6 expects static imports and emits the following warnings. In general, I believe it is bad practice to include dynamic requirements in a module - even while a dynamic import is being considered in ES.

I have replaced my project path with '/path/to/project' in the traces below:
```
WARNING in /path/to/project/index.js
47:15-60 Critical dependency: the request of a dependency is an expression
 @ ./node_modules/fastify-swagger/index.js
 @ ./es6/lib/app/index.js
 @ ./es6/index.js
```
I'll send a PR for this.

scriptjs:


scriptjs:
You can't build a path dynamically so removing the package.json make your title explicit which it should be in any case.

delvedor:
I'd like to leave the default title to the `package.json` title property.
You can accomplish the same result if you use `JSON.parse(fs.readFileSync(path.resolve(...)))` inside a try catch block.

index.js
```diff
@@ -45,7 +44,7 @@ function fastifySwagger (fastify, opts, next) {
     } else {
       swaggerObject.info = {
         version: '1.0.0',
-        title: require(path.join(__dirname, 'package.json')).name || ''
+        title: ''
```
mcollina: can you use fs.readFileSync to read this?

scriptjs: Yup, that's an alternative, I update

scriptjs:
Added tests for case where no info object is provided.

scriptjs:
Reverted tests but added sensible tests for title and version in info object as well as test for default swagger version.

scriptjs:
Are we all good here?

mcollina:
Yes!

PR #25, opened 2018-01-21
Updates plugin-meta

lependu:
see: https://github.com/fastify/fastify-plugin/issues/17#issuecomment-355972265


mcollina:
Thanks!

PR #28, opened 2018-02-13
add route consumes to swagger documentation

cortopy:
In swagger documentation the `consumes` property can be specified at top level or on a per-route basis. 

This PR adds the `consumes` property if specified at route level. This is important for swagger-ui too insofar as users can "try out" endpoints only with the content-type specified per route


mcollina:

PR #29, opened 2018-02-16
Parse request form params

cortopy:
https://github.com/fastify/fastify-formbody parses request body regardless of whether content-type is json or application/x-www-form-urlencoded
However, swagger has different schemas for json and application/x-www-form-urlencoded, multipart/form-data

This PR distinguishes between the two

### DIFF
--- a/index.js
+++ b/index.js
@@ -192,18 +192,14 @@ function getBodyParams (parameters, body) {
 
 function getFormParams (parameters, body) {
   const formParamsSchema = body.properties
-  if (!formParamsSchema) {
-
-  } else {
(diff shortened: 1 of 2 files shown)


cortopy:
@mcollina tests fail for node 6 and 4 as I use object destructuring. Are those versions actually supported?

mcollina:
Yes they are.
Il giorno lun 19 feb 2018 alle 15:00 Juan J. Jimenez-Anca <
notifications@github.com> ha scritto:

cortopy:
done

delvedor:
Landed in `v0.5.0` :)

PR #30, opened 2018-02-23
fastify-swagger static mode

PavelPolyakov:
hi @delvedor,

As we've discussed here is the update for the `fastify-swagger` which allows to serve the static swagger specification.

Please review the overall form, after your comments and if everything is fine I'd update the tests/readme/module version etc.


PavelPolyakov:
Hello @delvedor, friendly reminder, could you please share your feedback?

Regards,

delvedor:
Hi!
I'm stuck at work and I can't review this right now, I'll do this asap, sorry for the delay.
In the meantime can you fix the test under Node v4 and remove the `package-lock.json` file?
Thanks.

PavelPolyakov:
@delvedor sure, sorry for disturbing. Take your time.

reg v4 - I know what is the issue, I'll fix it along with the further work finalising if you approve the overall concept. 

`package-lock.json` - I thought `npm` recommends to commit it. I'll remove it as well.

PavelPolyakov:
I've updated README and removed `package-lock.json`.

Though I think you might want to reconsider this, since [here](https://docs.npmjs.com/files/package-lock.json) they recommend to commit it. Or maybe I miss something.

mcollina:
@PavelPolyakov here is why: https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514.

PavelPolyakov:
hello @delvedor, is there any chance this will be merge soon? I'd like to use it in the project.

static.js
```diff
@@ -0,0 +1,86 @@
...
+    )
+    const extName = path.extname(opts.specification.path).toLowerCase()
+    assert(
+      ['.yaml', '.json'].indexOf(extName) !== -1,
+      "specification.path extension name is not supported, should be one from ['.yaml', '.json']"
+    )
+
+    // read
+    const source = fs.readFileSync(
+      `${process.cwd()}/${opts.specification.path}`,
```
delvedor: This is not cross platform, you should use `path.resolve` instead.
Same above :)

tap-snapshots/test-static.js-TAP.test.js
```diff
@@ -0,0 +1,32 @@
+/* IMPORTANT
```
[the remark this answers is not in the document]
PavelPolyakov: I think it's common practice to commit snapshots, so they are used during the assertion

static.js
PavelPolyakov: updated, thanks

index.js
```diff
@@ -1,291 +1,22 @@
...
-// so '/user/:id' is not valid.
-// This function converts the url in a swagger compliant url string
-// => '/user/{id}'
-function formatParamUrl (url) {
-  var start = url.indexOf('/:')
-  if (start === -1) return url
 
-  var end = url.indexOf('/', ++start)
+  // by default the mode is dynamic, as plugin initially was developed
+  opts.mode = opts.mode && ['dynamic', 'static'].indexOf(opts.mode.toLowerCase()) !== -1 ? opts.mode.toLowerCase() : 'dynamic'
```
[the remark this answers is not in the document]
PavelPolyakov: updated, thanks

tap-snapshots/test-static.js-TAP.test.js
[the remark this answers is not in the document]
mcollina: Yes, this is needed.

delvedor:
Can you fix linting rules?

PavelPolyakov:
Sure, doing this, "it works on my machine" for some reason:
image

delvedor:
In master we are using standard v11 :D

PavelPolyakov:
all done :)

README.md
```diff
@@ -86,53 +86,79 @@ fastify.ready(err => {
...
-#### swagger options
-Calling `fastify.swagger` will return to you a JSON object representing your api, if you pass `{ yaml: true }` to `fastify.swagger`, it will return you a yaml string.
+  *All the above parameters are optional.*  
+  You can use all the properties of the [swagger specification](https://swagger.io/specification/), if you find anything missing, please open an issue or a pr!
+
+  Example of the `fastify-swagger` usage in the `dynamic` mode is available [here](examples/dynamic.js).
+<a name="mode.static"></a>
 
-If you pass `{ exposeRoute: true }` the plugin will expose the documentation with the following apis:
+* `static` mode should be configured explicitly. In this mode `fastify-swagger` serves given specification, you should craft it yourselfe.
```
mcollina: I would prefer if the dynamic and static modes are made subheadings `###` instead of `*`.

README.md
```diff
@@ -86,53 +86,79 @@ fastify.ready(err => {
...
+  ```js
+  {
+    mode: 'static',
+    specification: {
+      type: 'file',
```
mcollina: are there other types?

index.js
```diff
@@ -1,291 +1,24 @@
...
+
+  switch (opts.mode) {
+    case 'static':
+      setup.static(fastify, opts, next)
+      break
+    case 'dynamic':
+      setup.dynamic(fastify, opts, next)
+      break
+    default:
+      throw new Error("unsupported mode, should be one of ['static', 'dynamic']")
```
mcollina: Can you call `next()` with this error instead?

static.js
```diff
@@ -0,0 +1,86 @@
...
+    )
+    assert(
+      fs.existsSync(path.resolve(opts.specification.path)),
+      `${opts.specification.path} does not exist`
+    )
+    const extName = path.extname(opts.specification.path).toLowerCase()
+    assert(
+      ['.yaml', '.json'].indexOf(extName) !== -1,
+      "specification.path extension name is not supported, should be one from ['.yaml', '.json']"
+    )
```
mcollina: Because of the asynchronous loading of fastify, all those asserts should call `next(err)`.

test/static.js
```diff
@@ -0,0 +1,73 @@
...
+      // console.log(res.payload)
+      try {
+        var payload = JSON.parse(res.payload)
+        t.matchSnapshot(payload)
+      } catch (error) {
+        t.fail(error)
+      }
+    }
+  )
+})
```
mcollina: Can you please add some tests around these validations?

README.md
PavelPolyakov: right now not, but I thought that it will be possible to provide `url` or `file`

mcollina: then let's not have a  `type` at all.

static.js
PavelPolyakov: This is a good catch, do you see (or know) if it is possible to achieve this with `assert`? (without me diving into the fastify core code)
Because in the other module, where I actually test the validation, I was forced to refactor it to:
```
  if (!opts.secret) { return next(new Error('fastify-keycloak: secret should be provided')); }
```

if I use `assert(false, next('error'))`, what happens with the exception which is thrown by `assert`?

image

mcollina: You can wrap your asserts with a try/catch block.

PavelPolyakov: yes, I've also tried this, but after thought that `if` and `next(new Error())` is better.

ok, thanks for your comments and feedback, I'll try to fix them later this day.

README.md
PavelPolyakov: done, but hierarchically it's `#####`

README.md
PavelPolyakov: done

index.js
PavelPolyakov: done, thanks

static.js
PavelPolyakov: refactored to `if` + `return next(new Error())`

test/static.js
PavelPolyakov: added additional test

PavelPolyakov:
all done v2

PavelPolyakov:
@delvedor 
sorry for pursuing you, I can imagine you have things to do :) but I really want to have it integrated to the one of the projects I'm working on currently

delvedor:
@PavelPolyakov thank you for all you work and sorry for the delay!!

PavelPolyakov:
np, thank you, waiting for the new npm then :)

PR #34, opened 2018-03-08
Add swagger2 props

cortopy:
This PR does three things mainly:

- Removes $id property from shared schemas. The problem is that `$id` is not accepted in swagger. There is $ref but that would require a much bigger work
- Creates a new file where open api specs can be defined for more accurate parsing. After working with fastify-swagger for a bit I have realised that there a lot of properties that may be ignored from the definition in routes. This file is a way to start parsing objects by *picking* using some sort of dictionary of references. If this work is accepted and continued, there won't be the need to to an `omit` of invalid properties like `$id`
- Use the new file to parse the response object. A few commits ago, I started using the `headers` property in the response object. The problem is that it was added twice, as there was no way to differentiate between the body object (fastify only defines the body) and the headers


cortopy:
@mcollina this PR is now broken. I can see a lot has changed since I first requested it with the dynamic and static configurations. Not possible to merge right now. I'll try again in future

PR #36, opened 2018-03-15
serving oauth2-redirect.html; /documentation => /documentation/;

PavelPolyakov:
Started to integrate my swagger plus OAuth security schema and met an issue with `fastify-swagger`.
Seems we are only serving `index.html`, but not the other file.

Plus, right now it's not actually possible to use it under the url `/documentation/` (it will serve html, but  will not be able to load `css` and `js`). Therefore I've changed the paths to the absolute ones.

It works both `/documentation` and `/documentation/` right now.

As per `oauth2-redirect.html` - starting to serve it.

~~seems to be good, but I did not check the full cycle yet.~~
I can confirm that the full cycle "works on my machine".

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ If you pass `{ exposeRoute: true }` during the registration the plugin will expo
 |-------|----------------|
 |`'/documentation/json'` | the json object representing the api  |
 |`'/documentation/yaml'` | the yaml object representing the api  |
-|`'/documentation'` | the swagger ui  |
+|`'/documentation/'` | the swagger ui  |
 
 <a name="swagger.options"></a>
 ### swagger options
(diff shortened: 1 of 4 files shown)


PavelPolyakov:
Ok, but there seem to be downside as well. Like if I host `/documentation/` but on production I reverse proxy the folder `/my-service` to the service root. 
So I access `/my-service/documentation/`. And this means that swagger UI will resolve from the root of the domain - not good.

Is there something like `resolveUrl` but for `css` and `js`? Or in which way we can solve it 🤔

PavelPolyakov:
At the end I think having one url to serve documentation is the most safest solution.
I've put redirect from `/documentation` to `/documentation/`. This should also work if server is served from the folder (`/service/documentation` will be redirected to `/service/documentation/`).

PavelPolyakov:
@delvedor friendly reminder, please review and release the new version if everything is fine

PR #37, opened 2018-03-23
server swagger-ui using the swagger-ui-dist module

PavelPolyakov:
I've checked that in the old (current) version there is a bug regarding oAuth.
Probably it was already solved in the original swagger-ui/dist.

Regarding the permissions - let me know if I should put them back. Since I don't see any reason to have them `0755` and `0644` are from the original repo.


mcollina:
How did you update swagger-ui? Can we use https://www.npmjs.com/package/swagger-ui-dist instead?

PavelPolyakov:
@mcollina 
I just copied over the files from the current master. 
Actually in the current state of PR I forgot to serve new png files. 

Yes, we can use `dist` as dependency, kind of I do it here:
https://github.com/PavelPolyakov/fastify-serve-swagger-ui/blob/master/src/index.js#L58

also it's possible to serve everything through `fastify-static`.

would you support integration with static + dist ?

PavelPolyakov:
done as `static` + `dist`.

in my previous PR there was a bug for redirecting `/documentation` to `/documentation/` when being served by the reverse proxy. So if you were calling `/upstream/documentation` it was redirecting you to `/documentation/`. Now should be fixed, tested with `fastify-http-proxy`.

since we are not storing static files anymore, I've removed the assertions which were testing the content.

this [hook](https://github.com/fastify/fastify-swagger/pull/37/files#diff-d640e5835a6b0c342cf705a06018c2b9R63) is a little bit strange as for me, since it will really listen for the all incoming requests. Let me know if you think there is better solution.

routes.js
```diff
@@ -43,27 +50,26 @@ function fastifySwagger (fastify, opts, next) {
...
-    } else {
-      return reply.code(404).send(new Error('Not found'))
+  // hijak swagger index.html response
+  fastify.addHook('onSend', (request, reply, payload, next) => {
+    if (
+      request.raw.originalUrl === `/documentation/` ||
+      request.raw.originalUrl === `/documentation/index.html`
+    ) {
+      reply.header('Content-Type', 'text/html; charset=UTF-8')
+      payload = files.index
```
mcollina: this should not be needed, maybe fastify-static has a bug. Why do you need it?

mcollina: I don't think these prefixes are correct. I can mount this plugin with a given `/swagger`  and it will add the ui as `/swagger/documentation/`.

A better approach would be to either have a `post-install` script or a `prepublish`  script where we do these changes and replacements. A prepublish might be better, because we can't guarantee that regexps are stable across versions. Why do we need those changes anyway?

PavelPolyakov: This changes needed because by default it `index.html` of the swagger ui leads to the pet store.
https://github.com/swagger-api/swagger-ui/blob/master/dist/index.html

> I can mount this plugin with a given /swagger and it will add the ui as /swagger/documentation/.
how would you do that? I think `prefix` option is not supported.

I've tried this together with `example/dynamic.js`:
```js
fastify.register(require('../index'), {
  swagger: {
    info: {
      title: 'Test swagger',
      version: '0.1.0'
    },
    host: 'localhost',
    schemes: ['http'],
    consumes: ['application/json'],
    produces: ['application/json']
  },
  exposeRoute: true,
  prefix: '/swagger'
})
```

the documentation appeared on `localhost:3000/documentation` still.

> a prepublish  script where we do these changes and replacements. 
you mean on prepublish we copy everything into our static folder and then serve it instead of the original `swagger-ui-dist`?

PavelPolyakov: Here is the new variant https://github.com/fastify/fastify-swagger/pull/39

PavelPolyakov:
closing since the other implementation won

PR #39, opened 2018-03-25
serve swagger-ui through the swagger-ui-dist; v2

PavelPolyakov:
this is another variant of https://github.com/fastify/fastify-swagger/pull/37 .
preparing the `static` folder before publishing the module (during the development needs to be done manually)


.npmignore
```diff
@@ -0,0 +1,53 @@
...
+*.swp
+
+package-lock.json
+
+# project specific
+*.tgz
+tap-snapshots
+test
+.travis.yml
+var
```
mcollina: Can you remove this file?

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "Generate Swagger files automatically for Fastify.",
   "main": "index.js",
   "scripts": {
-    "test": "standard && tap test/*.js"
+    "prepare:swagger-ui": "node var/npm/prepare-swagger-ui",
```
mcollina: can you use `prepublish`  instead? https://docs.npmjs.com/misc/scripts.

routes.js
```diff
@@ -43,28 +28,15 @@ function fastifySwagger (fastify, opts, next) {
...
-    handler: (request, reply) => reply.redirect(request.raw.url + '/')
+    handler: (request, reply) => reply.redirect('./documentation/')
   })
 
-  fastify.route({
-    url: '/documentation/:file',
-    method: 'GET',
-    schema: { hide: true },
-    handler: sendStaticFiles
+  // server swagger-ui with the help of fastify-static
```
mcollina: typo, it should be "serve swagger-ui with the help of fastify-static".

.npmignore
PavelPolyakov: yes, I can, but, I wanted `static` to be ignored by git while be included in the npm. I see it's only possible with adding `.npmignore`.

plus I think that it's not really needed to include tests and other stuff into the package itself.

routes.js
PavelPolyakov: fixed

package.json
PavelPolyakov: fixed

.npmignore
mcollina: Oh right. Ok.

routes.js
```diff
@@ -43,28 +28,15 @@ function fastifySwagger (fastify, opts, next) {
     url: '/documentation',
     method: 'GET',
     schema: { hide: true },
-    handler: (request, reply) => reply.redirect(request.raw.url + '/')
+    handler: (request, reply) => reply.redirect('./documentation/')
```
delvedor: Why the dot in `'./documentation/'`?

PavelPolyakov: so my case was the next, I serve my service under the reverse-proxy, like:
`/service` points to the service itself, let's say `localhost:3000`.

With the previous version, when I was visiting `/service/documentation` it was redirecting me to the `/documentation/`. With the current version it works fine.

Here how I've checked it:
```js
'use strict'

const fastify = require('fastify')()
const proxy = require('fastify-http-proxy')

fastify.register(require('../index'), {
  mode: 'static',
  specification: {
    path: './examples/example-static-specification.yaml'
  },
  exposeRoute: true
})

fastify.register(proxy, {
  upstream: 'http://localhost:3000/',
  prefix: '/upstream/' // optional
})

fastify.listen(3000, err => {
  if (err) throw err
  console.log('listening')
})
```
(put to the script into `examples`, install `fastify-http-proxy`, open `http://localhost:3000/upstream/documentation`)

var/npm/prepare-swagger-ui.js
```diff
@@ -0,0 +1,39 @@
+const fs = require('fs')
```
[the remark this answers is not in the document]
PavelPolyakov: I like to have the root clean, since usually in the modern web project there are already a lot of dot files out there.

So usually I have:
- `src` (for the code, which sometimes get compiled into the `dist`)
- `var` (for all the various files such as build scripts, images for the readme, etc.)
- maybe `__tests__`

And that's all.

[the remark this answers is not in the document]
delvedor: Since it is just one file you can leave it inside the root folder, it is not our common practice use the `var` folder and this could lead to confusion :)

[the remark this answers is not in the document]
PavelPolyakov: np, done

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "Generate Swagger files automatically for Fastify.",
   "main": "index.js",
   "scripts": {
-    "test": "standard && tap test/*.js"
+    "prepare:swagger-ui": "node prepare-swagger-ui",
+    "test": "standard && tap test/*.js",
+    "prepublish": "npm run prepare:swagger-ui && npm test"
```
mcollina: npm test is not needed here.

package.json
```diff
@@ -22,12 +24,16 @@
...
     "fastify": "^1.1.1",
+    "fs-extra": "^5.0.0",
     "standard": "^11.0.0",
     "swagger-parser": "^4.0.2",
-    "tap": "^11.1.0"
+    "tap": "^11.1.0",
+    "swagger-ui-dist": "3.13.0"
   },
   "dependencies": {
+    "fastify-http-proxy": "^0.2.1",
```
mcollina: I guess fastify-http-proxy should be in `devDependencies`.

PavelPolyakov: sorry `fastify-http-proxy` should not be here at all, removed

package.json
PavelPolyakov: removed

routes.js
mcollina: I think this is ok.

PavelPolyakov:
can this be released? 🤗

PavelPolyakov:
Thanks!

PR #41, opened 2018-03-29
correctly resolving static folder

PavelPolyakov:
I'm sorry, there was a 🐞 .

root was resolving into the root of the project folder, not the module folder. I've realised it only after I've installed v0.7.0 into my project.

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const fp = require('fastify-plugin')
-const resolve = require('path').resolve
+const path = require('path')
 
 function fastifySwagger (fastify, opts, next) {
   fastify.route({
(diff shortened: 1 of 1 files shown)


mcollina:

PavelPolyakov:
yes, I thought about this, but this is kind of unclear for me how to reproduce this in the tests, since `npm test` is running under the `fastify-swagger` root (there are already tests for the `staitc` mode). And `node examples/static-file.js` was also running from the module directory.

So everything was resolved correctly while you were working with the `fastify-swagger`.

mcollina:
I think you can use https://nodejs.org/api/process.html#process_process_chdir_directory to change the current directory.

routes.js
```diff
@@ -33,7 +33,7 @@ function fastifySwagger (fastify, opts, next) {
 
   // serve swagger-ui with the help of fastify-static
   fastify.register(require('fastify-static'), {
-    root: resolve('./static'),
+    root: resolve(__dirname, 'static'),
```
mcollina: you might want to use `path.join(__dirname, 'static')` instead. I'm good with no test in that case.

PavelPolyakov: done, thank you for the suggestion

PR #56, opened 2018-04-22
adding the ability to change swagger url

doron2402:
- Adding the ability to customize the swagger end-point by passing `routePrefix` 
- use `/documentation` by default

@mcollina

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![Greenkeeper badge](https://badges.greenkeeper.io/fastify/fastify-swagger.svg)](https://greenkeeper.io/)
 [![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat)](http://standardjs.com/)  [![Build Status](https://travis-ci.org/fastify/fastify-swagger.svg?branch=master)](https://travis-ci.org/fastify/fastify-swagger)
 
-[Swagger](https://swagger.io/) documentation generator for Fastify.  
+[Swagger](https://swagger.io/) documentation generator for Fastify.
 It uses the schemas you declare in your routes to generate a swagger compliant doc.
 
 <a name="install"></a>
@@ -26,6 +26,7 @@ fastify.register(require('fastify-swagger'), {
       description: 'testing the fastify swagger api',
       version: '0.1.0'
     },
+    routePrefix: '/documentation',
     host: 'localhost',
     schemes: ['http'],
     consumes: ['application/json'],
@@ -115,7 +116,7 @@ fastify.ready(err => {
   }
   ```
 
-  *All the above parameters are optional.*  
+  *All the above parameters are optional.*
   You can use all the properties of the [swagger specification](https://swagger.io/specification/), if you find anything missing, please open an issue or a pr!
 
   Example of the `fastify-swagger` usage in the `dynamic` mode is available [here](examples/dynamic.js).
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -26,6 +26,7 @@ fastify.register(require('fastify-swagger'), {
       description: 'testing the fastify swagger api',
       version: '0.1.0'
     },
+    routePrefix: '/api-docs',
```
mcollina: can you set this to the default?

dynamic.js
```diff
@@ -27,7 +27,8 @@ module.exports = function (fastify, opts, next) {
   const security = opts.swagger.security || null
 
   if (opts.exposeRoute === true) {
-    fastify.register(require('./routes'))
+    const routePrefix = opts.routePrefix || '/documentation'
+    fastify.register(require('./routes'), { routePrefix })
```
mcollina: you can pass `{ prefix }` instead. That's more aligned with the fastify way.

routes.js
```diff
@@ -4,8 +4,9 @@ const fp = require('fastify-plugin')
 const path = require('path')
 
 function fastifySwagger (fastify, opts, next) {
+  const routePrefix = opts.routePrefix || '/documentation'
   fastify.route({
-    url: '/documentation/json',
+    url: `${routePrefix}/json`,
```
mcollina: if you pass `{ prefix }`  as an option, then you just have to specify `/json` as your route. See https://github.com/fastify/fastify/blob/master/docs/Plugins.md#plugin-options. Same for every other route.

routes.js
```diff
@@ -25,16 +26,16 @@ function fastifySwagger (fastify, opts, next) {
   })
 
   fastify.route({
-    url: '/documentation',
+    url: `${routePrefix}`,
     method: 'GET',
     schema: { hide: true },
-    handler: (request, reply) => reply.redirect('./documentation/')
+    handler: (request, reply) => reply.redirect(`.${routePrefix}/`)
```
mcollina: I'm not 100% sure how `reply.redirect()` works with the prefix. Open an issue on fastify if it does not work.

README.md
doron2402: 👍

dynamic.js
```diff
@@ -27,7 +27,8 @@ module.exports = function (fastify, opts, next) {
   const security = opts.swagger.security || null
 
   if (opts.exposeRoute === true) {
-    fastify.register(require('./routes'))
+    const prefix = opts.routePrefix || '/api-docs'
+    fastify.register(require('./routes'), { prefix })
```
doron2402: @mcollina 
I did not have time today to dive in why this one is not working. but for some reason when passing `{ prefix: '/api-docs' }` tests are failing and I had to manually use the `opts.prefix`

doron2402: I was expecting that the `prefix` attribute will automatically add a prefix and I wouldn't need to add `opts.prefix` to the routes.

doron2402: I think I got it, I was expecting that `prefix` will be used inside of a plugin (same as a route).

doron2402:
Hi @mcollina 
Should we wait until we fix the bug with the `prefix` before merging this pr?
thanks.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fastify-swagger",
-  "version": "0.8.3",
+  "version": "0.8.4",
```
mcollina: the maintainer bumps the version number. Can you please revert this change?

routes.js
```diff
@@ -5,7 +5,7 @@ const path = require('path')
 
 function fastifySwagger (fastify, opts, next) {
   fastify.route({
-    url: '/documentation/json',
+    url: `${opts.prefix}/json`,
```
mcollina: this should not be needed.

static.js
```diff
@@ -35,7 +35,8 @@ module.exports = function (fastify, opts, next) {
   fastify.decorate('swagger', swagger)
 
   if (opts.exposeRoute === true) {
-    fastify.register(require('./routes'))
+    const prefix = opts.routePrefix || '/api-docs'
```
mcollina: the default should be `/documentation`.

mcollina:
I've left some notes on the PR.

> Should we wait until we fix the bug with the prefix before merging this pr?


Which one?

routes.js
doron2402: seems like route prefixing doesn't work when wrapping the plugin with `fastify-plugin`

routes.js
```diff
@@ -5,7 +5,7 @@ const path = require('path')
 
 function fastifySwagger (fastify, opts, next) {
   fastify.route({
-    url: `${opts.prefix}/json`,
+    url: '/json',
```
doron2402: @mcollina Seems like when using `fastify-plugin` it's ignoring the route prefix

mcollina: there is no need to use fastify-plugin in ./routes.js

doron2402: Got it, I was confused why it was used there to begin with.

I'll take it out.

doron2402:
ok @mcollina i think it's finally ready :)
Thanks for the comments 👍

PR #67, opened 2018-05-07
Fix: redirect to index

### LINKED ISSUE #65: Wrong redirect
The redirection should be made with an absolute url and redirect to `/documentation/index.html`

https://github.com/fastify/fastify-swagger/blob/master/routes.js#L10

allevo:
Fix #65

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -7,7 +7,7 @@ function fastifySwagger (fastify, opts, next) {
     url: '/',
     method: 'GET',
     schema: { hide: true },
-    handler: (request, reply) => reply.redirect(`${opts.prefix}/index.html`)
+    handler: (request, reply) => reply.redirect(`${fastify.basePath}/index.html`)
   })
 
   fastify.route({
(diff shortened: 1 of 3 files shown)


routes.js
```diff
@@ -7,7 +7,7 @@ function fastifySwagger (fastify, opts, next) {
     url: '/',
     method: 'GET',
     schema: { hide: true },
-    handler: (request, reply) => reply.redirect(`.${opts.prefix}/`)
+    handler: (request, reply) => reply.redirect(`${opts.prefix}/index.html`)
```
mcollina: I don't think that `opts.prefix` here is correct. Can you add a test that loads this within another plugin, so that this is exposed as `/v1/documentation`? I think it should use [`fastify.basePath`](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#basepath).

allevo: nice catch!! fixed

mcollina:
Feel free to release it to land and release it to npm as minor!

allevo:
@mcollina I have no right to merge this PR

mcollina:
Now you can :D

Issue #69, opened 2018-05-11
Wrong redirect


sjorsjes:
v0.10.1 - There stil seems to be a bug when using `routePrefix: '/'` even after this fix (https://github.com/fastify/fastify-swagger/issues/65) from @allevo. It redirects the browser to `index.html` instead of `http://domain/index.html`

allevo:
Hi!
  sorry for the inconvenient.

With the following test, which redirection are you expecting ?

```
test('with routePrefix: \'/\' should redirect /', t => {
  t.plan(4)
  const fastify = Fastify()

  const opts = JSON.parse(JSON.stringify(swaggerInfo))
  opts.routePrefix = '/'
  fastify.register(fastifySwagger, opts)

  fastify.get('/foo', () => {})

  fastify.inject({
    method: 'GET',
    url: '/'
  }, (err, res) => {
    t.error(err)
    t.strictEqual(res.statusCode, 302)
    t.strictEqual(res.headers['location'], '????????????????')
    t.is(typeof res.payload, 'string')
  })
})
```

The current implementation (`0.10.1`) redirects to `//index.html` (with double slashes)

sjorsjes:
The current implementation forces my browser to go to ‘index.html’ instead of ‘domain/index.html’. The fix I suggested allows for a single slash and makes the browser go to the right path. You can test it by setting the ‘routePrefix’ option to ‘/‘

PR #70, opened 2018-05-11
Fix: Wrong redirect #69

sjorsjes:
Fix for https://github.com/fastify/fastify-swagger/issues/69

If the trailing slash does not exist, a '/' is appended to the end


sjorsjes:
I'm not able to checkout/commit at the moment. Making tests through the web editor is not really something i want to do :/

mcollina:
cc @allevo

PR #71, opened 2018-05-11
Fix #69

### LINKED ISSUE #69: Wrong redirect

allevo:
cc @sjorsjes


sjorsjes:
Much better solution then what I came up with 👍

Issue #59, opened 2018-04-27
generate fastify code  from swagger definition ?


seriousme:
Hi,

Since fastify already uses JSON-schema I thought it might be an idea to try to generate fastify code from a swagger doc so one can use a Design-First approach as well.  
Either using [swagger codegen](https://github.com/swagger-api/swagger-codegen), a fastify specific tool or fully dynamically by using JSON path queries on a swagger definition (using a fastify plugin ?)

Has anybody already tried this ?

Swagger codegen currently only generates Express code for NodeJS

Kind regards,
Hans

mcollina:
@seriousme I think it would be a very nice thing to have indeed. I'm not a huge fan of code generators as the generated code tends to get out of sync very quickly.

I'd be very happy to review anything, if you are willing to work on the topic.

seriousme:
I'm aware of the sync problem ;-)
The idea is to use the swagger ```operationId``` to link the generated code to the rest of the code.

I'll see what I can come up with :-)

seriousme:
Ok, I'm making some progress.
However I run into a schema issue , probably easy to solve, but I just don't see it :-(

Swaggerfiles tend to put their schema in the `definitions` property at top level and then in the path definitions refer to them as e.g. `"$ref": "#/definitions/Pet"`  (from the [petstore example](http://petstore.swagger.io/v2/swagger.json))
```json
  "parameters": [
          {
            "in": "body",
            "name": "body",
            "description": "Pet object that needs to be added to the store",
            "required": true,
            "schema": {
              "$ref": "#/definitions/Pet"
            }
          }
        ],
```
refering to:
```json
"definitions"{
  "Pet": {
      "type": "object",
      "required": ["name", "photoUrls"],
      "properties": {
        "id": {
          "type": "integer",
          "format": "int64"
        },
        "category": {
          "$ref": "#/definitions/Category"
        },
        "name": {
          "type": "string",
          "example": "doggie"
        }
     }
  }
}
```
In fastify I try to create one shared schema (the definitions section) and then have the schema's per route refer to that.

```javascript
const fastify = require("fastify")();

fastify.addSchema({
  $id: "defs",
  definitions: {
    amount: {
      type: "number",
      required: true
    }
  }
});

const schema = {
  querystring: {
    amount: { $ref: "defs#/definitions/amount" }
  }
};
// Declare a route
fastify.get("/", { schema }, async (request, reply) => {
  //console.log(request);
  return { hello: "world" };
});

[...]

PR #75, opened 2018-05-19
Added reference to fastify-swaggergen to README.md

seriousme:
as discussed in #59


mcollina:
@delvedor I think it's useful to document this here as well.

mcollina:
Thanks!!

PR #86, opened 2018-07-14
Add missing tags and externalDocs top-level fields

### LINKED ISSUE #85: Support `tags` and `externalDocs` in dynamic mode
As per OpenAPI 2.0 specification, you are missing 2 top-level fields described here https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields:
- `tags`
- `externalDocs`

Because you are reconstructing `swaggerObject` from `opts.swagger` in `dynamic.js` even if those fields are given, they are stripped in the output.

I'm filing this issue because I found that `tags` was missing, but there may be more fields that are lacking deeper down, I didn't check.

ofavre:
Fixes #85.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,18 @@ fastify.register(require('fastify-swagger'), {
       description: 'testing the fastify swagger api',
       version: '0.1.0'
     },
+    externalDocs: {
+      url: 'https://swagger.io',
+      description: 'Find more info here'
+    },
     host: 'localhost',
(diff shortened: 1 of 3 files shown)


ofavre:
Updated.

mcollina:
This is now conflicting with master. can you please rebase?

ofavre:
Rebased.

Issue #87, opened 2018-07-14
Support openapi 3.0 for dynamic configuration


SkeLLLa:
Seems underlying module `swagger-parser` already supports new openapi v3 specifications. So I think it would be good if this module will support it as well. 

The main idea just to make https://github.com/fastify/fastify-swagger/blob/master/dynamic.js#L59 configurable through options.

However it will also require to refactor some things like `securityDefinitions`, because in 3.0 they moved to other sections of config file.

I'm not sure that fastify schemas are compatible with openapi v3 specs.

sjorsjes:
Does this include Bearer auth aswell? https://swagger.io/docs/specification/authentication/bearer-authentication/

PR #88, opened 2018-07-14
WIP: feat(openapi): add initial openapi support

SkeLLLa:
***WIP***
This PR adds initial openapi v3.X.X support for dynamic mode.

Related: #87

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -6,7 +6,6 @@ const yaml = require('js-yaml')
 
 module.exports = function (fastify, opts, next) {
   fastify.decorate('swagger', swagger)
-
   const routes = []
 
   fastify.addHook('onRoute', (routeOptions) => {
@@ -18,6 +17,7 @@ module.exports = function (fastify, opts, next) {
   opts.swagger = opts.swagger || {}
 
   const openApiVersion = opts.openapi || null
+  const isOAS3 = Boolean(openApiVersion)
   const info = opts.swagger.info || null
   const host = opts.swagger.host || null
   const schemes = opts.swagger.schemes || null
@@ -32,7 +32,7 @@ module.exports = function (fastify, opts, next) {
 
   if (opts.exposeRoute === true) {
     const prefix = opts.routePrefix || '/documentation'
-    fastify.register(require('./routes'), { prefix })
+    fastify.register(require('./routes'), {prefix})
   }
 
   const cache = {
@@ -59,7 +59,7 @@ module.exports = function (fastify, opts, next) {
     // Base swagger info
     // this info is displayed in the swagger file
     // in the same order as here
-    if (openApiVersion) {
+    if (isOAS3) {
       swaggerObject.openapi = openApiVersion
     } else {
       swaggerObject.swagger = '2.0'
@@ -77,18 +77,29 @@ module.exports = function (fastify, opts, next) {
     if (basePath) swaggerObject.basePath = basePath
     if (consumes) swaggerObject.consumes = consumes
     if (produces) swaggerObject.produces = produces
-    if (definitions) swaggerObject.definitions = definitions
-    if (securityDefinitions) {
-      swaggerObject.securityDefinitions = securityDefinitions
-    }
-    if (security) {
-      swaggerObject.security = security
-    }
-    if (components) {
-      swaggerObject.components = components
-    }
-    if (servers) {
-      swaggerObject.servers = servers
+    if (securityDefinitions) swaggerObject.securityDefinitions = securityDefinitions
+    if (security) swaggerObject.security = security
+    if (components) swaggerObject.components = components
+    if (servers) swaggerObject.servers = servers
+
+    if (opts && opts.addFastifySchemas) {
+      const schemas = fastify._schemas.store
+      const schemaKeys = Object.keys(schemas)
+      let schemasDst = swaggerObject.definitions
+      if (isOAS3) {
+        swaggerObject.components.schemas = definitions || {}
+        schemasDst = swaggerObject.components.schemas
+      } else {
+        swaggerObject.definitions = definitions || {}
+        schemasDst = swaggerObject.definitions
+      }
+
+      for (var schemaKey of schemaKeys) {
+        const schema = Object.assign({}, schemas[schemaKey])
+        const id = schema.$id
+        delete schema.$id
+        schemasDst[id] = schema
+      }
     }
 
     swaggerObject.paths = {}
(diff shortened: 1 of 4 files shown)


SkeLLLa:
However it's still should be checked. 
[] Check and implement new body format https://swagger.io/docs/specification/describing-request-body/

mcollina:
Can you please add unit test for this?

mcollina:
Overall, I do not know enough of OpenAPI 3 or have enough time to come up with a solution to those two problems. 

Would you like to send a PR? Or describe what is your proposed solution? Please assume we know very little about OpenAPI 3.

mcollina:
I’m all for A)

SkeLLLa:
Ok. Then I'll try to do the trick body params location and add some test.
But for the future I'd like to have at least `oneOf` in fastify.

dynamic.js
```diff
@@ -70,12 +77,29 @@ module.exports = function (fastify, opts, next) {
...
+        schemasDst = swaggerObject.components.schemas
+      } else {
+        swaggerObject.definitions = definitions || {}
+        schemasDst = swaggerObject.definitions
+      }
+
+      for (var schemaKey of schemaKeys) {
+        const schema = Object.assign({}, schemas[schemaKey])
+        const id = schema.$id
+        delete schema.$id
```
mcollina: Can you add a comment why this is done?

dynamic.js
```diff
@@ -123,11 +147,25 @@ module.exports = function (fastify, opts, next) {
...
+        }
+
+        if (schema.requestBody) {
+          const content = schema.requestBody.content
+          parameters.push(content)
+          // const consumesAllFormOnly =
+          //     consumesFormOnly(schema) || consumesFormOnly(swaggerObject)
+          // consumesAllFormOnly
+          //   ? getFormParams(parameters, schema.body, !!swaggerObject.openapi)
+          //   : getBodyParams(parameters, schema.body, !!swaggerObject.openapi)
```
mcollina: why is this block commented out?

SkeLLLa:
As for today status is:
- [ ] Add more OAS3 features support
- [ ] Request body now correctly generated for OAS3
- [ ] Added feature to support generating openapi definitions not only from specific field, but also from fastify schemas (for both swagger 2.0 and OAS3)
- [ ] Support `oneOf` and `anyOf` for OAS3
- [ ] Update docs in README
- [ ] Add some tests

dynamic.js
SkeLLLa: This will be removed. I've made correct implementation within body processing.

dynamic.js
SkeLLLa: Yeah. Added comment. 
So main idea - is to extract fastify schemas to swagger definitions if proper option will be set.

dynamic.js
```diff
@@ -70,12 +77,30 @@ module.exports = function (fastify, opts, next) {
...
-    if (security) {
-      swaggerObject.security = security
+    if (securityDefinitions) swaggerObject.securityDefinitions = securityDefinitions
+    if (security) swaggerObject.security = security
+    if (components) swaggerObject.components = components
+    if (servers) swaggerObject.servers = servers
+
+    if (opts && opts.addFastifySchemas) {
+      // extract definitions from fastify schemas
+      const schemas = fastify._schemas.store
```
mcollina: I would prefer if you did not access this here. Can you send a PR to fastify to get an iterator over the schema store? Or a  `getAllSchemas()` method or accessor?

SkeLLLa: Good point. I've thought about this. For now it will remain just for tests and after PR to fastify I'll replace it with appropriate method.

SkeLLLa:
I've done initial implementation and added some tests for basic features. 
However I don't like how code looks like. Probably it would be better even to write separate plugin. Or do some major refactoring moving all generation-related functions to separate modules for v2 and v3 with common interface.

mcollina:
Good work! I'm sorry for the delay in reviewing this, but I'm currently traveling and I didn't had time. 
I would actually prefer to have the common interface as you have described, or maybe have some internal utilities for the common parts and two separate files like static and dynamic.

SkeLLLa:
@mcollina I've finished some work in separate repo. I'll commit it soon and post link here.

SkeLLLa:
@mcollina Here it is: https://github.com/SkeLLLa/fasify-oas.
Basic functionality should work. Also I've tried to keep it compatible with current version. So in order to migrate to new OAS3. It tries to extract as much info from swagger 2.0 config as possible and reuse it in it's own config.
I'll continue adding tests to be sure that it covers OAS as much as possible.

SkeLLLa:
@mcollina I've finished pre-release version. Added some tests and made it maximum compatible with this plugin. So in order to generate OpenApi docs you'll need just to replace `fastify-swagger` with `fastify-oas`. However it doesn't support so called 'static' mode, because there's already good plugin that does the thing: https://github.com/seriousme/fastify-openapi-glue

Also it supports new features that were introduced in OpenApi 3.0 like servers, components, etc.

A bit later I'll also test a new openapi plugin on real project. And after that it will be released.

As for integration - I don't know if it will be good idea to mix all the things in one plugin. There are lot's of features introduced in openapi 3.0 and more to come. So it will be quiet hard to maintain 4 things at once in one repo. I mean static, dynamic, swagger 2.0 and openapi 3.0. And as separate modules I think it will be easier to keep code clean.

So may be it will be good idea if this plugin will become some sort of meta-plugin that will depend on separate modules for openapi, swagger, static, dynamic. All of them will share some common interface and it will allow to select proper sub-plugin depending on user configuration and providing it via common namespace.

The code is available at:
https://gitlab.com/m03geek/fastify-oas
https://github.com/SkeLLLa/fastify-oas

mcollina:
Awesome, so I guess we can close this, right?

SkeLLLa:
@mcollina, yes. let's close this.

Marsup:
Why close this one? I don't see a future for this plugin if it doesn't support latest specs and it feels like a good start.

SkeLLLa:
@Marsup It's implemented as separate plugins: 
https://github.com/seriousme/fastify-openapi-glue - if you need generate Fastify routes from openapi schema
https://gitlab.com/m03geek/fastify-oas - if you need to generate openapi schema from Fastify routes.
It has compatible interface with this one, so if you're looking for migration it should be very easy.

Marsup:
I understood that and switched to fastify-oas, but that's basically a fork that would have no reason to exist if this one had up-to-date spec support, right?

PR #92, opened 2018-08-01
introducing postProcessor for static mode

PavelPolyakov:
🤚,

I suggest to introduce an option of `postProcessor` to be able dynamically adjust the static file, for example based on the environment you run your application on.

My use case - I have auth integrated into the swagger specification:
```
oAuth2Code:
      type: oauth2
      description: OAuth
      flows:
        authorizationCode:
          authorizationUrl: https://localhost:9000/and/it/goes/further
          tokenUrl: https://localhost:9000/and/it/goes/further
          scopes:
            adminLogin: "Login as admin"
```

These urls are different for `localhost`/`development`/`staging`/`whatever`.

Having a `postProcessor` option will allow me to inject them based on the `NODE_ENV` and `config`, at the same time keeping something as the default value.

Regards,

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,11 +138,17 @@ fastify.ready(err => {
   {
     mode: 'static',
     specification: {
-      path: './examples/example-static-specification.yaml'
+      path: './examples/example-static-specification.yaml',
+      postProcessor: function(swaggerObject) {
+        return swaggerObject
+      }
     }
   }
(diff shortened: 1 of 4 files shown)


mcollina:
Good work!

PavelPolyakov:
Thank you, you're the best!

PR #94, opened 2018-08-07
enum configures in path params is missing

BigMurry:
For the following schema configure:
```
{
  params: {
    type: 'object',
    properties: {
      enumKey: {type: 'string', enum: ['enum1', 'enum2']}
   }
  }
}
```
the expected swagger json result should not omit the **enum: ["enum1", "enum2"]** part


mcollina:
Would you mind adding a unit test?

PR #98, opened 2018-08-16
feature: support explicitly passed document #97

### LINKED ISSUE #97: Support passing swagger definition in static mode
Currently, in `static` mode, there is no way to pass a *content* of the file directly. You need to pass the path to the file instead.

## Current behavior
```
mode: 'static',
specification: {
    path: './examples/example-static-specification.yaml'
},
```

## New feature request
```
mode: 'static',
specification: {
    doc: {
        ...real swagger document... (content of the specification file)
    }
},
```

## Why?
We use `fastify-swagger` to provide a fastify integration with NestJS (and Swagger module exposed by `@nestjs/swagger`). Unluckily, we cannot fully support this platform since there is no way to pass swagger definition's object directly (you always have to pass a path to the file in static mode). Would it be a big deal to provide such possibility in the future?

kamilmysliwiec:
resolves #97 by adding `document` property which accepts valid JavaScript object (already loaded & parsed swagger specification)

### DIFF
--- a/static.js
+++ b/static.js
@@ -10,33 +10,40 @@ module.exports = function (fastify, opts, next) {
 
   let swaggerObject = {}
 
-  if (!opts.specification.path) return next(new Error('specification.path is missing, should be path to the file'))
-  if (typeof opts.specification.path !== 'string') return next(new Error('specification.path is not a string'))
+  if (!opts.specification.path && !opts.specification.document) {
+    return next(new Error('both specification.path and specification.document are missing, should be path to the file or swagger document spec'))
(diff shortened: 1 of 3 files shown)


mcollina:
Can you add a unit test for this feature?

kamilmysliwiec:
Not very familiar with this tooling 😅

mcollina:
You should duplicate https://github.com/fastify/fastify-swagger/blob/master/test/static.js#L43-L73 and tweak to test the new functionality.

kamilmysliwiec:
Done

PR #105, opened 2018-09-26
Added platform specific slash

cemremengu:
Testing on windows is hard.

This looks ugly so feel free to close if you think it's unnecessary

### DIFF
--- a/test/route.js
+++ b/test/route.js
@@ -8,6 +8,7 @@ const yaml = require('js-yaml')
 const fastifySwagger = require('../index')
 
 const resolve = require('path').resolve
+const path = require('path').join
 const readFileSync = require('fs').readFileSync
 
 const swaggerInfo = {
@@ -183,7 +184,7 @@ test('/documentation should redirect to /documentation/index.html', t => {
   }, (err, res) => {
     t.error(err)
     t.strictEqual(res.statusCode, 302)
-    t.strictEqual(res.headers['location'], '/documentation/index.html')
+    t.strictEqual(res.headers['location'], `${path('/', 'documentation', 'index.html')}`)
     t.is(typeof res.payload, 'string')
   })
 })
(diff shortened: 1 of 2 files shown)


cemremengu:
image

test/route.js
```diff
@@ -183,7 +184,7 @@ test('/documentation should redirect to /documentation/index.html', t => {
   }, (err, res) => {
     t.error(err)
     t.strictEqual(res.statusCode, 302)
-    t.strictEqual(res.headers['location'], '/documentation/index.html')
+    t.strictEqual(res.headers['location'], `${path('/', 'documentation', 'index.html')}`)
```
delvedor: Did you test this on Windows?
I'm not sure it will work.

Ps: usually we wait for at least one *LGTM* before merging a pr :)

mcollina:
I think this has to be reverted. the snapshot file is automatically generated by TAP, so it should not updated manually.

Also, headers always use `/`.

test/route.js
cemremengu: Right I rushed a bit, apologies for that and yes I am on windows :)

Why do you think that?

delvedor: Nevermind, read @mcollina's comment :D

cemremengu:
I am sorry for this. Any other ideas how we can run tests on windows?

mcollina:
what error are you getting?

mcollina:
https://github.com/fastify/fastify-swagger/blob/master/routes.js#L10 is a bug and it should not use `path.join`. The library is in fact broken on windows.

cemremengu:
Changing to `handler: (request, reply) => reply.redirect(`${fastify.basePath}/index.html`)` doubles the `/` in one test, rest is fine

```
  with routePrefix: '/' should redirect to /index.html
  not ok should be equal
    +++ found
    --- wanted
    -/index.html
    +//index.html
```

mcollina:
that should be fixed

Il giorno ven 28 set 2018 alle 09:59 Cemre Mengu <notifications@github.com>
ha scritto:

cemremengu:
#110

PR #110, opened 2018-09-28
Fix incorrect basePath concat

cemremengu:


mcollina:
Can you also do a patch release? Remember to fill in https://github.com/fastify/fastify-swagger/releases as well.

cemremengu:
Could you bump the `package.json` directly or should I send a PR?

cemremengu:
Nevermind I was able to commit directly

PR #116, opened 2018-10-11
Remove type property when value is null

### LINKED ISSUE #115: Declare 204 with type null
When I declare the scheme as below

```
response: {
        204: {
          description: 'Succesful response',
        },
      },
```

I get the following stacktrace:

```
Error: undefined unsupported
    at build (/Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/fast-json-stringify/index.js:94:13)
    at getValidatorForStatusCodeSchema (/Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/fastify/lib/validation.js:13:10)
    at /Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/fastify/lib/validation.js:19:21
    at Array.reduce (<anonymous>)
    at getResponseSchema (/Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/fastify/lib/validation.js:18:22)
    at build (/Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/fastify/lib/validation.js:58:31)
    at afterRouteAdded (/Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/fastify/fastify.js:607:9)
    at Object._encapsulateThreeParam (/Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/avvio/boot.js:387:7)
    at Boot.callWithCbOrNextTick (/Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/avvio/boot.js:311:5)
    at Boot._after (/Users/guilherme.gregio/Documents/easynvest/easynvest.preferencies/node_modules/avvio/boot.js:215:26)
``` 

When I try use type: 'null' 
```
response: {
        204: {
          type: 'null',
          description: 'Succesful response',
        },
      },
```

But the result is "Unknown Type: null" in model's section on swagger doc

guilhermegregio:
This PR closes #115

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -279,6 +279,9 @@ function genResponse (response) {
 
   Object.keys(response).forEach(key => {
     if (response[key].type) {
+      if (response[key].type === 'null') {
+        delete response[key].type
+      }
       var rsp = response[key]
       var description = response[key].description
       var headers = response[key].headers


dynamic.js
```diff
@@ -279,6 +279,9 @@ function genResponse (response) {
 
   Object.keys(response).forEach(key => {
     if (response[key].type) {
```
cemremengu: I am not really familiar with swagger but just peeking at examples I don't think `type` is mandatory. If `type` is not required, maybe we can remove `if (response[key].type)` all together to solve the problem?

Any thoughts?

guilhermegregio: The problem when remove the type in schema like bellow 
```
schema: {
      response: {
        204: {
          description: 'Succesful response',
        },
      },
    },
```
It shows an error on fastify validation 'fast-json-stringify'

Any thoughts?

cemremengu:
@StarpTech @mcollina Is this related to here? https://github.com/fastify/fast-json-stringify/blob/master/index.js#L69 

Maybe `undefined` should be added?

mcollina:
Can you add a unit test?

@cemremengu this should be ok to land afterwards.

StarpTech:
@cemremengu @mcollina I checked it and this change isn't needed please check my comment https://github.com/fastify/fastify-swagger/issues/115#issuecomment-429083320 We could use the special "nullable" type when someone wants to handle a null value as a response type.

cemremengu:
@StarpTech I agree, that would be a better solution

cemremengu:
Closing this due to inactivity (and probably not needed). Feel free to reopen .

Issue #121, opened 2018-11-06
Corrupt file attachments


Stezido:
We encountered an issue where a file download triggered via the Content-Disposition header will result in a link to download the file, but the file will be corrupt. A similiar issue is also documented in following swagger-ui issue:

https://github.com/swagger-api/swagger-ui/issues/4638

In our case we want to download a tar-file. So the content-type of the response is ```application/gzip```.  We want to add the download-endpoint to the swagger-documentation but if we do that the downloaded file is corrupt.

In the https://github.com/swagger-api/swagger-ui/issues/4638 Issue the solution is to upgrade to Swagger UI 3. Is there a workaround in fastify-swagger?

cemremengu:
Isn't swagger ui same as the dependency `"swagger-ui-dist": "3.19.5"` or am I missing something ?

cemremengu:
@Stezido Can you try `0.15.2` to see if that works for you ?

Stezido:
> Isn't swagger ui same as the dependency `"swagger-ui-dist": "3.19.5"` or am I missing something ?

You are right fastify-swagger is already using swagger ui 3. This issue is caused by ajv the validator of fastify. For someone who is running in a similar issue i summed up my story.

The Swagger docu describes a "[response that returns a file](https://swagger.io/docs/specification/describing-responses/#response-that-returns-a-file )"
My response schema looks like this for a gzip-file:
```
response: {
       200: {
          type : "string",
          format: "binary",
      },
 },
```
The ajv validator offers following **types**: 
number, integer, string, boolean, array, object or null
 and following **formats** for type "string": 
date, date-time, uri, email, hostname, ipv4, ipv6, regex

So there is a type "string" but no format "binary"!

Only defining the type as "string" or "object" doesn't make it because the file downloaded will be corrupt

add a byte/binary format to ajv via addFormat would be my guess how to fix it, but I couldn't add a new format because fast-json-stringify uses it's own instance of ajv internally.
[This issue](https://github.com/fastify/fast-json-stringify/issues/115) describe a problem of fast-json-stringify I also ran into. The issue was solved in version 1.8.0. of fast-json-stringify by setting the option unknownFormats: "ignore" to the intern ajv instance. So updating fastify made it for me.

cemremengu:
@mcollina should we update fastify range of the plugin for this?

PR #125, opened 2018-11-19
Update fastify range

cemremengu:
See #121. 

1.8.0 of stringify is introduced here https://github.com/fastify/fastify/commit/4741dc6e64ee8516cb0b40656bbae374fa823d7e#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,6 +23,6 @@ function fastifySwagger (fastify, opts, next) {
 }
 
 module.exports = fp(fastifySwagger, {
-  fastify: '>=0.39.0',
+  fastify: '>=1.12.1',
   name: 'fastify-swagger'
 })


index.js
```diff
@@ -23,6 +23,6 @@ function fastifySwagger (fastify, opts, next) {
 }
 
 module.exports = fp(fastifySwagger, {
-  fastify: '>=0.39.0',
+  fastify: '>=1.12.1',
```
mcollina: I fear this might fail in v2, as it depends on fastify-static.

cemremengu: Should we bump `fastify-static` to the same version as well?

mcollina: no, not really. fastify-static should be `^1.0.0` ideally.

cemremengu:
Closing this in favor of #130 and 1.x approach discussed

PR #130, opened 2018-12-05
Fixes for fastify v2

cemremengu:
Tests fail but I couldn't figure out. The redirect for `/documentation` -> `/documentation/index.html` does not return any payload for some reason

Failing test: https://github.com/fastify/fastify-swagger/blob/master/test/route.js#L286

Failing line: https://github.com/fastify/fastify-swagger/blob/master/routes.js#L14

Any ideas?


mcollina:
@cemremengu this should fix the test!

mcollina:
what’s the status of this?

cemremengu:
Was waiting for #137 for covering more functionality. Should we land?

mcollina:
I think so.

PR #136, opened 2018-12-16
Branching off to 1.x

cemremengu:


cemremengu:
We should mark this merge commit as 1.x branch

PR #137, opened 2018-12-18
Support static files references

kromol:


mcollina:
Thanks for your contribution!

Linting is failing:

```
standard: Use JavaScript Standard Style (https://standardjs.com)
standard: Run `standard --fix` to automatically fix some problems.
  /home/travis/build/fastify/fastify-swagger/routes.js:46:65: Extra semicolon.
  /home/travis/build/fastify/fastify-swagger/routes.js:49:31: Extra semicolon.
  /home/travis/build/fastify/fastify-swagger/routes.js:50:15: Extra semicolon.
  /home/travis/build/fastify/fastify-swagger/routes.js:55:41: Extra semicolon.
  /home/travis/build/fastify/fastify-swagger/static.js:54:42: Unexpected trailing comma.
```

Also, can you add a unit test?

kromol:
Yep, I'll fix it. This is still work in progress. I just opened PR so you can track the progress from the beginning :)

kromol:
@mcollina are you OK with using `fastify@next` for now and updating it to released version later?
Currently there are issues with router which are fixed in fastify 2.

cemremengu:
@kromol I was waiting for this PR to land actually but if it is really necessary I will bump the version

mcollina:
> Currently there are issues with router which are fixed in fastify 2.

Which issue?
You can target https://github.com/fastify/fastify-swagger/tree/fastify-next, which is the update to the new version of fastify.

kromol:
I meant this issue - https://github.com/fastify/fastify/issues/1338

Currently the problem is that plugin will have to serve static files from two different locations:
1. Internal `static` directory with UI stuff
2. Directory with yaml/json documentation files <-- this should be dynamic

With current setup I initially was thinking about adding additional regex route to handle only yaml and json files and do not break other static stuff handled by `fastify-static`. Since it did not work I started to think about using `fastify@next` to overcome routing issues...

But I just realized that another way might be to put everything served by `fastify-static` under it's own prefix. It seems better option from my perspective since it separates UI stuff from other plugin endpoints and should work with current dependencies. I will give it a try today.

kromol:
Happy New Year!
@mcollina pls review the PR when you get a chance

test/route.js
```diff
@@ -418,3 +420,32 @@ test('/documentation2/json route (overwrite)', t => {
...
+  fastify.register(fastifySwagger, swaggerInfo)
+
+  fastify.inject({
+    method: 'GET',
+    url: '/documentation/swagger-ui.js'
+  }, (err, res) => {
+    t.error(err)
+    t.strictEqual(res.statusCode, 404)
+  })
+})
```
mcollina: Can you please add a test that checks that the Fastify not found handler is called? This is likely to fail/conflict on Fastify v2, and we need the test to make sure the behavior is consistent.

routes.js
```diff
@@ -38,7 +42,35 @@ function fastifySwagger (fastify, opts, next) {
...
+
+        const ext = path.extname(file).toLocaleLowerCase()
+        if (!fs.existsSync(filePath) || ['.yaml', '.json'].indexOf(ext) === -1) {
+          return reply.code(404).send()
+        }
+
+        const contentType = ext === '.yaml' ? 'application/x-yaml' : 'application/json'
+        reply
+          .type(contentType)
+          .send(fs.readFileSync(filePath))
```
mcollina: I think it's better if you switch to fastify-static `reply.sendFile()` method. You can just add the decorator by passing `{ serve: false }` when registering fastify-static.

You'll need to register two instances of fastify-static, one with `{ serve: false }` to use inside this function, and one with `{ decorateReply: false }` to serve the swagger-ui. The two instances should not conflict/error.

test/route.js
kromol: Done

routes.js
kromol: Done

kromol:
Rebased my branch on the latest `master`, but did not squash commits yet for easier review

README.md
```diff
@@ -143,14 +143,17 @@ fastify.ready(err => {
...
+      baseDir: '/path/to/external/spec/files/location',
+    },
   }
   ```
   Example of the `fastify-swagger` usage in the `static` mode is available [here](examples/static-file.js).
 
   `specification.postProcessor` parameter is optional. It allows you to change your swagger object on the fly (for example - based on the environment). It accepts `swaggerObject` - basically a javascript object which was parsed from your `yaml` or `json` file and should return a swagger object.
 
+  `specification.baseDir` allows to specify directory where all spec files which are included into the main one using `$ref` live. 
+  By default it's the directory where main spec file is located. Should be absolute path without trailing slash.
```
cemremengu: allows specifying the directory where all spec files that are included in the main one using `$ref` will be located .  By default, this is the directory where the main spec file is located. Provided value should be an absolute path **without** trailing slash.

cemremengu: > Provided value should be an absolute path without trailing slash.

Why don't we trim the trailing slash if exists, instead of relying on user to know this?

kromol: @cemremengu yep, you are right, I forgot to implement it. Will fix in the next commit. And on the other hand keeping this as part of documentation (even with trimming trailing slashes) would make it more obvious for the users how to use it (IMO)

kromol: Done

kromol: Done

routes.js
```diff
@@ -38,7 +41,29 @@ function fastifySwagger (fastify, opts, next) {
 
   // serve swagger-ui with the help of fastify-static
   fastify.register(require('fastify-static'), {
-    root: path.join(__dirname, 'static')
+    root: path.join(__dirname, 'static'),
+    prefix: staticPrefix,
+    decorateReply: false
+  })
+
+  fastify.register(require('fastify-static'), {
```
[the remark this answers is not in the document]
kromol: Yep, done.

Issue #139, opened 2018-12-28
Fix types to support all fields in static mode


4a57:
There are incomplete typing for static mode - missing `routePrefix` and `exposeRoute`

4a57:
Yes. Do I need some rights to push new branch?

mcollina:
Just fork the repository and push to your fork.

climba03003:
Resolved by #140

PR #140, opened 2018-12-29
Fix types for static mode (#139)

4a57:
PR for issue #139

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,20 +5,20 @@ import * as SwaggerSchema from 'swagger-schema-official';
 declare namespace fastifySwagger {
   interface FastifySwaggerOptions {
     mode?: 'static' | 'dynamic';
-  }
-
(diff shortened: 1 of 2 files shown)


mcollina:
Can you please add a unit test for these new types?

4a57:
done

index.d.ts
```diff
@@ -5,20 +5,19 @@ import * as SwaggerSchema from 'swagger-schema-official';
...
     routePrefix?: string;
     /**
-     *  To expose the documentation api
+     * To expose the documentation api
      * @default false
      */
     exposeRoute?: boolean;
+  }
+
+  interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
```
mcollina: I think `mode?: 'dynamic';` should be kept here. Why did you remove it?

4a57: Oh, right... Fixed

PR #141, opened 2019-01-09
This bugfix is related to #137 PR

kromol:
Since we added prefix `/static` to separate swagger ui static files from other files we need to make sure that client wlll not use `/static` prefix for other calls (like for retrieving swagger.yaml)

### DIFF
--- a/prepare-swagger-ui.js
+++ b/prepare-swagger-ui.js
@@ -32,7 +32,7 @@ const newIndex = fs.readFileSync(resolve('./static/index.html'), 'utf8')
   }`)
   .replace(
     /url: "(.*)",/,
-    `url: resolveUrl('./json'),
+    `url: resolveUrl('./json').replace('static/json', 'json'),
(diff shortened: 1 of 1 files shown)


mcollina:
How can i test/verify this?

kromol:
I don't know other way than checking it in test project.
I just tried to use it in the project I'm working on, faced this issue and applied a fix.
I can create simple test project later if needed.

mcollina:
There is https://github.com/fastify/fastify-swagger/tree/master/examples. Can you explain what's failing?

kromol:
We moved swagger ui under `static/` prefix. So URL in browser by default looks like this: `/documentation/static/index.html`. `index.html` resolved this URL `./json` in: 

```
const ui = SwaggerUIBundle({
        url: resolveUrl('./json')
        ....
```

to `/documentation/static/json` that is wrong, because `static/` prefix serves only swagger ui.

PR #149, opened 2019-02-05
Bugfix missing `operationid` and `produces` route property

pgaubatz:
This is an optional field: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-5


mcollina:
Can you add a unit test?

pgaubatz:
> Can you add a unit test?

done! :)

mcollina:
Thanks!

PR #155, opened 2019-02-26
feat(routeSchema types): Adds routeSchema types

karlos1337:
Allow using route schema swagger properties


karlos1337:
Fix https://github.com/fastify/fastify-swagger/issues/156

Issue #156, opened 2019-03-01
Fastify-Swagger introduction to a TYPESCRIPT proj doen't allow swagger details


panbhatt:
While importing fastify-swagger into a fastify (ts) based project, when we are defining the schema of any method, we are restricted because of the RouteSchema defined in the fastify.d.ts. 
For Swagger we need to expand this to include other properties like documenation,tags,summary,security. 
Creating a pull reques to acheive this.

karlos1337:
Fixed here https://github.com/fastify/fastify-swagger/pull/155 , pending merge

PR #157, opened 2019-03-01
Extension to Route Schema for Swagger. 

panbhatt:
Fix for issue https://github.com/fastify/fastify-swagger/issues/156 

To allow Fastify TS project to define a broader schema, so that SWAGGER can take proper values to be shown in the UI.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -54,6 +54,13 @@ declare module 'fastify' {
       }
     ) => SwaggerSchema.Spec;
   }
+  
+    interface RouteSchema {
+    description?: String,
+    tags?: String[],
+    summary?: String,
+    security?: any[]
+  }
 }
 
 declare let fastifySwagger: fastify.Plugin<
--- a/index.d.ts
+++ b/index.d.ts
@@ -55,11 +55,11 @@ declare module 'fastify' {
     ) => SwaggerSchema.Spec;
   }
   
-    interface RouteSchema {
-    description?: String,
-    tags?: String[],
-    summary?: String,
-    security?: any[]
+  interface RouteSchema {
+    description?: string;
+    tags?: string[];
+    summary?: string;
+    security?: any[];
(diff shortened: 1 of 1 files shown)


cemremengu:
Can you add a unit test? Also fix the formatting to match rest of the code?

panbhatt:
The test is already covered for this on test/swagger.js at line no 468/471/472. Security can be different for each application so it doesnt' apply here. 
Commiting the formatting changes.

index.d.ts
```diff
@@ -54,6 +54,13 @@ declare module 'fastify' {
       }
     ) => SwaggerSchema.Spec;
   }
+  
+  interface RouteSchema {
+    description?: string;
+    tags?: string[];
+    summary?: string;
+    security?: any[];
```
Ethan-Arrowood: This shouldn't be any; what are the actual objects types of the `security` list?

panbhatt: It is being kept as ANY, to accommodate wide variety of use cases. It can be a JWT token or some specific name that is specific to the implementation. I think not keeping it tight to a specific we are opening the possibilities to a wide variety of use cases. Everyone can implement it in their own way. Let me know your thoughts, we can also delete it. if not being applied here.

ab18556: I am new to it so I might be wrong, but shouldn't all this stuff be based on this?

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9d9c2ff7419b5e05bf91b29f765beb8d146ee987/types/swagger-schema-official/index.d.ts

karlos1337:
Already fixed here https://github.com/fastify/fastify-swagger/pull/157 , pending merge

index.d.ts
Ethan-Arrowood: Yes if this type could infer the values provided by swagger that would be best

PR #169, opened 2019-04-09
HTTP2 Support added

RSWilli:
changed types to support http2
added a http2 types testfile


mcollina:
@Ethan-Arrowood can you take a look?

PR #185, opened 2019-05-31
fix(typings): RouteSchema

Bapen1k:
fastify-swagger doesn't inherit the original RouteSchema and therefore the types don't work correctly. I don't know how to fix it

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -56,7 +56,7 @@ declare module 'fastify' {
     ) => SwaggerSchema.Spec;
   }
 
-  interface RouteSchema {
+  interface RouteSchema extends fastify.RouteSchema {
     hide?: boolean;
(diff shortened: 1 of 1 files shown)


Eomm:
> fastify-swagger doesn't inherit the original RouteSchema and therefore the types don't work correctly. I don't know how to fix it

Would you mind to open an issue where you describe better the problem and how replicate it since the CI is failing?

Bapen1k:
> > fastify-swagger doesn't inherit the original RouteSchema and therefore the types don't work correctly. I don't know how to fix it
> 
> Would you mind to open an issue where you describe better the problem and how replicate it since the CI is failing?

ok, i created issue [https://github.com/fastify/fastify-swagger/issues/186](url)

PR #215, opened 2019-11-29
Fix typing: array definition

hKaspy:
The old typing throws TypeError when using `security: []` definition in endpoint schema:

image

Workaround for that is to use `[{}]` as seen with `const b`, but that in turn confuses Swagger UI which renders it as secured, pre-authenticated endpoint:

image

This new typing fixes that:

image
image

#### Checklist

### DIFF
--- a/http2-types.test.ts
+++ b/http2-types.test.ts
@@ -26,6 +26,14 @@ app.put('/some-route/:id', {
     }
   }, (req, reply) => {});
 
+app.get('/public/route', {
+  schema: {
+    description: 'returns 200 OK',
+    summary: 'qwerty',
+    security: []
+  }
+}, (req, reply) => {});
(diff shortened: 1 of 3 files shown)


mcollina:
can you add a test?
Il 29 nov 2019, 14:41 +0100, Jan Kašpar <notifications@github.com>, ha scritto:

hKaspy:
Only for the case of using `security: []` in js. I didn't see any tests regarding the typings file. Would that be ok?

mcollina:
This file: https://github.com/fastify/fastify-swagger/blob/master/types.test.ts.

hKaspy:
Must have been blind. Will add tests on Monday, 4th, when I get back from the weekend ;-)

PR #218, opened 2020-01-09
Add "exposeRoute:true" in usage example

olafkrueger:
The default of `exposeRoute` is `false` which leads to a 404 error:

{"message":"Route GET:/documentation not found","error":"Not Found","statusCode":404}

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ fastify.register(require('fastify-swagger'), {
         in: 'header'
       }
     }
-  }
+  },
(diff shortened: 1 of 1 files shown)


erayalakese:
I'm still having 404 error. `exposeRoute` is `true`

{"message":"Route GET:/documentation not found","error":"Not Found","statusCode":404}

mcollina:
I'd recommend opening a fresh issue with a full example.

Issue #178, opened 2019-05-08
How can I change collectionFormat?


foisonocean:
Hi guys, is there a way to change collectionFormat of querystrings?

lddubeau:
Yes. Here's an example of code I'm using:

```
    server.route({
      method: "GET",
      url,
      schema: {
        querystring: {
          type: "object",
          required: ["fields"],
          additionalProperties: false,
          properties: {
            fields: {
              type: "array",
              items: {
                type: "string",
              },
              collectionFormat: "multi", // <== HERE IT IS!
              minItems: 1,
            },
          },
        },
      },
      handler: [...],
    });
```

Prior to that change the Swagger UI would produce the `fields` parameter like this `fields=a,b`. After the change, it produces `fields=a&fields=b`, which is what the default query parser wants to produce an array.

AFAICT, changing the `collectionFormat` only changes Swagger's behavior. Your query parser **also** has to support the format you want, and the default parser supports parameter repetition (`multi`, like I'm using above), not other formats. So if you use the default query parser you pretty much must specify `collectionFormat: "multi"` for all your arrays or it won't work. If you want another format, you need to replace the default parser with another one (like `qs`, for instance), or use your own custom parsing code.

mcollina:
Nice @lddubeau! Would you like to send a PR with the above as docs and maybe an example?

lddubeau:
@mcollina  [PR](https://github.com/fastify/fastify-swagger/pull/223) submitted.

climba03003:
Resolved by #223

PR #223, opened 2020-02-06
docs: document how to set OA-specific parameter options

lddubeau:
As requested in https://github.com/fastify/fastify-swagger/issues/178#issuecomment-582627553

This documents how to set Open API-specific parameter options in a schema used for parsing query strings.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,8 @@ fastify.route({
           minItems: 1,
           //
           // Note that this is an Open API version 2 configuration option.  The
-          // options changed in version 3.
+          // options changed in version 3. The plugin currently only supports
+          // version 2 of Open API.
           //
           // Put `collectionFormat` on the same property which you are defining
           // as an array of values. (i.e. `collectionFormat` should be a sibling
--- a/README.md
+++ b/README.md
@@ -214,6 +214,50 @@ fastify.register(require('fastify-swagger'), {
 ### swagger options
 Calling `fastify.swagger` will return to you a JSON object representing your api, if you pass `{ yaml: true }` to `fastify.swagger`, it will return you a yaml string.
 
+### Open API (OA) Parameter Options
+
+Note: OA's terminology differs from Fastify's. OA uses the term "parameter" to refer to those parts of a request that in [Fastify's validation documentation](https://www.fastify.io/docs/latest/Validation-and-Serialization/#validation) are called "querystring", "params", "headers".
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -214,6 +214,50 @@ fastify.register(require('fastify-swagger'), {
...
+      properties: {
+        fields: {
+          type: 'array',
+          items: {
+            type: 'string'
+          },
+          minItems: 1,
+          //
+          // Note that this is an Open API version 2 configuration option.  The
+          // options changed in version 3.
```
mcollina: Good work! Which is the option for v3?

lddubeau: I decided to mention it without going into the gory details so that *if* someone tries it in v3 the note gives them fair warning. I decided against going into more details mainly because:

1. I think the plugin currently does not let us use v3. Mind you, the way I came to this conclusion was that I shoved the v3 options meant to do the same thing as I was doing in v2 in my configuration and it did nothing good. Then I eyeballed the code and [saw](https://github.com/fastify/fastify-swagger/blob/70461eeaa5ea7290b1549e98b7c03f3ac1511a15/dynamic.js#L62):

   ```
    swaggerObject.swagger = '2.0'
   ```

   I scanned the code and did not see how this specific field might be overwritten somehow. Hmm... actually now that I review this, I'm thinking that having Swagger run in v3 mode would require more than changing the value of this field. As I recall, `swagger` is not a valid field in v3 at all. When running in v3 mode, it should set another field, the `openapi` field I believe.

   Also, I did not find mention in the plugin documentation as to how to select an Open API version. Switching versions is a big deal because v3 is not backward compatible with v2. So the lack of documentation was (for me) another indication that probably switching versions is not possible right now.
   
2. Going into the details of how it works in v3 would require more explanation than I think is warranted for the issue at hand here. For one thing, in v3 the configuration which is equivalent to v2's `collectionFormat: "multi"` *is the default handling in v3*. (The default in v2 was `collectionFormat: "csv"`.) So in fact, a lack of further specification in v3 gives a behavior which is compatible with the default query parser. Stated explicitly in v3's options, v3's default is becomes: `style: "form", explode: true`. Ok, how do these two interact?

   Basically, I'm trying to avoid turning the README into a tutorial on how to use Open API.

mcollina: Great writeup, can you mention that this module supports v2 and not v3 in the comment?

lddubeau: Done.

Issue #224, opened 2020-02-12
Querystring not working with shared schema


eriott:
## 🐛 Bug Report

Generating of swagger docs fails when I use shared schema (replace-way) in querystring.

## To Reproduce

1. Register fastify-swagger plugin.
1. Register any shared schema with id.
1. Create a route and use shared schema in querystring definition.
1. Run app and go to docs page.
1. An error happens:
 ```js
  TypeError: Cannot create property 'name' on string 'q'
    at Object.keys.forEach.prop (/home/regina/Projects/test-fastify-swagger/node_modules/fastify-swagger/dynamic.js:220:16)
    at Array.forEach (<anonymous>)
    at getQueryParams (/home/regina/Projects/test-fastify-swagger/node_modules/fastify-swagger/dynamic.js:217:22)
    at Object.swagger (/home/regina/Projects/test-fastify-swagger/node_modules/fastify-swagger/dynamic.js:144:11)
    at Object.handler (/home/regina/Projects/test-fastify-swagger/node_modules/fastify-swagger/routes.js:28:26)
    at preHandlerCallback (/home/regina/Projects/test-fastify-swagger/node_modules/fastify/lib/handleRequest.js:111:30)
    at preValidationCallback (/home/regina/Projects/test-fastify-swagger/node_modules/fastify/lib/handleRequest.js:100:5)
    at handler (/home/regina/Projects/test-fastify-swagger/node_modules/fastify/lib/handleRequest.js:69:5)
    at handleRequest (/home/regina/Projects/test-fastify-swagger/node_modules/fastify/lib/handleRequest.js:18:5)
    at onRunMiddlewares (/home/regina/Projects/test-fastify-swagger/node_modules/fastify/lib/middleware.js:22:5)

 ```

Use this code to reproduce error:

```js
const fastify = require('fastify')({
  bodyLimit: 1024 * 1024 * 2.5,
  ignoreTrailingSlash: true,
  logger: true
})

fastify.register(require('fastify-swagger'), {
  routePrefix: '/docs',
  exposeRoute: true,
  swagger: {
    info: {
      title: 'Test swagger',
      description: 'testing the fastify swagger api',
      version: '0.1.0'
    },
    host: 'localhost',
    schemes: ['http'],
    consumes: ['application/json'],
    produces: ['application/json'],
  }
})

fastify.addSchema({
  $id: 'query-schema',
  type: 'object',
  properties: {
    param1: {
      type: 'number'
    },
    param2: {
[...]

PR #226, opened 2020-02-13
Fixes validation when using shared schemas

leorossi:
It checks if `querystring` is a string rather than an object, then finds the related fastify schema (if exists), and use that object to generate swagger doc.

Fixes: #224

#### Checklist

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -5,6 +5,8 @@ const path = require('path')
 const yaml = require('js-yaml')
 
 module.exports = function (fastify, opts, next) {
+  const fastifySchemas = fastify.getSchemas()
+
   fastify.decorate('swagger', swagger)
 
   const routes = []
@@ -141,7 +143,7 @@ module.exports = function (fastify, opts, next) {
         }
 
         if (schema.querystring) {
-          getQueryParams(parameters, schema.querystring)
+          getQueryParams(parameters, schema.querystring, fastifySchemas)
         }
 
         if (schema.body) {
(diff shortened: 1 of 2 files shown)


dynamic.js
```diff
@@ -202,7 +204,13 @@ function consumesFormOnly (schema) {
   )
 }
 
-function getQueryParams (parameters, query) {
+function getQueryParams (parameters, query, fastifySchemas) {
+  if (typeof query === 'string' && query.match(/#$/)) {
```
mcollina: Would we have this problem also for the body if it is a string?

SkeLLLa: Yes, it will be the problem with body, params, headers and all the stuff that accepts schemas as references.

dynamic.js
```diff
@@ -5,6 +5,8 @@ const path = require('path')
 const yaml = require('js-yaml')
 
 module.exports = function (fastify, opts, next) {
+  const fastifySchemas = fastify.getSchemas()
```
SkeLLLa: I have some concerns about using `fastify.getSchemas()`, cause they will return schemes only from "current" scope. So in this fix will only support those schemes which are defined through fastify plugin.

In order to support all cases you'll probably need to get them recursively from all contexts. See https://github.com/SkeLLLa/fastify-oas/blob/master/lib/openapi/index.js#L5 for example.

SkeLLLa:
I think this issue was resolved in fastify/fastify#2108.

PR #233, opened 2020-03-31
Ensure paths start with a leading slash

### LINKED ISSUE #232: basePath is stripped from paths
## 🐛 Bug Report

When the `basePath` is set to `/`, the leading `/` is removed from paths.

## To Reproduce

```js
const Fastify = require("fastify");

const app = Fastify({
  logger: {
    level: 'info'
  }
});

app.register(require("fastify-swagger"), {
  routePrefix: "/spec",
  exposeRoute: true,
  swagger: {
    basePath: '/'
  }
});

app.route({
  method: 'GET',
  url: '/foo/bar',
  handler (request, reply) {
    reply.send()
  }
})

app.listen(8080, '127.0.0.1', (err, address) => {
  if (err) {
    app.log.error(err);
    process.exit(1);
  }
});

```

This is the generated spec:
```js
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.0",
    "title": "fastify-swagger"
  },
  "basePath": "/",
  "paths": {
    "foo/bar": {
      "get": {
        "responses": {
          "200": {
            "description": "Default Response"
          }
        }
      }
    }
  }
}
```
This is not valid

## Expected behavior

I expected the following spec:

```js
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.0",
    "title": "fastify-swagger"
  },
  "basePath": "/",
  "paths": {
    "/foo/bar": {
      "get": {
        "responses": {
          "200": {
            "description": "Default Response"
          }
        }
      }
    }
  }
}
```

(The same happens if the `basePath` is set to `/foo`, the path will be `/bar`)

## Environment

- *node version*: 10.19.0
- *fastify version*: 2.13.0
- *os*: Windows, Linux

---

This seems to be intentional, am I misusing the library?

Nepoxx:
Fixes #232 

#### Checklist

- [x] run `npm run test`
- [ ] `npm run benchmark`

I personally would remove the whole thing about stripping the basePath from paths, but since that would be a breaking change I decided against it.

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -96,9 +96,12 @@ module.exports = function (fastify, opts, next) {
       const schema = transform
         ? transform(route.schema)
         : route.schema
-      const path = route.url.startsWith(basePath)
+      let path = route.url.startsWith(basePath)
         ? route.url.replace(basePath, '')
         : route.url
+      if (!path.startsWith('/')) {
+        path = '/' + path
+      }
       const url = formatParamUrl(path)
 
       const swaggerRoute = swaggerObject.paths[url] || {}
--- a/test/swagger.js
+++ b/test/swagger.js
@@ -634,3 +634,46 @@ test('basePath support with prefix', t => {
     t.ok(swaggerObject.paths['/endpoint'])
   })
 })
+
+test('basePath ensure leading slash', t => {
+  t.plan(3)
+  const fastify = Fastify()
+
+  fastify.register(fastifySwagger, {
+    swagger: Object.assign({}, swaggerInfo.swagger, {
+      basePath: '/'
+    })
+  })
+
+  fastify.get('/endpoint', {}, () => {})
+
+  fastify.ready(err => {
+    t.error(err)
+
+    const swaggerObject = fastify.swagger()
+    t.notOk(swaggerObject.paths.endpoint)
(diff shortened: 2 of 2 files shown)


dynamic.js
```diff
@@ -96,9 +96,12 @@ module.exports = function (fastify, opts, next) {
       const schema = transform
         ? transform(route.schema)
         : route.schema
-      const path = route.url.startsWith(basePath)
+      let path = route.url.startsWith(basePath)
         ? route.url.replace(basePath, '')
         : route.url
+      if (!path.startsWith('/')) {
+        path = '/' + path
```
Eomm, line 103: I think this is correct, but do we know who is removing this slash?

Nepoxx: Yes, the line right above with
```js
route.url.replace(basePath, '')
```
if `basePath` is defined as `/`, then it will be removed everywhere.

Nepoxx:
Would you suggest mentioning the fact that the prefix is stripped from the paths in the README? That was not immediately clear to me.

andyburke:
I'm having an issue related to this: basePath is being removed from my routes, even though it shouldn't be.

Why is basePath being stripped from the urls? I am not sure what the benefit of that is supposed to be? At the least, it seems like this should be either gated behind a setting, or have a setting to opt out?

PR #235, opened 2020-04-23
WIP: Add support for multiple dynamic configs

gwn:
Sometimes it may make sense to generate slightly different
documentations based on the same routes by relying on the
`transform` option.

My specific scenario involves serving both a full and a limited
versions of the documentation. Here is my config:
```
app.register(fastifySwagger, [
    {
        routePrefix: '/limited-documentation-for-3rd-parties',
        transform: schema => ({...schema, hide: !schema.public}),
        exposeRoute: true,
    },
    {
        routePrefix: '/protected-full-documentation-for-our-clients',
        exposeRoute: true,
    }
])

```

The feature only applies to the `dynamic` mode and is backwards
compatible.

If anybody shows any interest to this feature; I can complete the PR by
sending tests, benchmarks and docs.

Feedbacks are welcome.

#### Checklist

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -13,6 +13,19 @@ module.exports = function (fastify, opts, next) {
     routes.push(routeOptions)
   })
 
+  opts = opts || [{}]
+
+  const setupFns =
+    opts.map((subopts, idx) => makeSetupFn(subopts, idx, fastify, routes, next))
+
+  function swagger (opts, idx) {
+    return setupFns[idx || 0](opts)
+  }
+
+  next()
+}
+
+function makeSetupFn (opts, idx, fastify, routes, next) {
   opts = opts || {}
 
   opts.swagger = opts.swagger || {}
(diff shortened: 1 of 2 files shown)


mcollina:
Go ahead. I would use an option instead of an array anyway.

gwn:
All right. Option makes sense, I'll consider it. I will get back. Thank you

dynamic.js
```diff
@@ -32,15 +45,15 @@ module.exports = function (fastify, opts, next) {
 
   if (opts.exposeRoute === true) {
     const prefix = opts.routePrefix || '/documentation'
-    fastify.register(require('./routes'), { prefix })
+    fastify.register(require('./routes'), { prefix, conf_idx: idx })
```
mcollina, line 48:
```suggestion
    fastify.register(require('./routes'), { prefix, confIdx: idx })
```

routes.js
```diff
@@ -25,7 +25,7 @@ function fastifySwagger (fastify, opts, next) {
     method: 'GET',
     schema: { hide: true },
     handler: function (req, reply) {
-      reply.send(fastify.swagger())
+      reply.send(fastify.swagger({}, opts.conf_idx))
```
mcollina, line 28:
```suggestion
      reply.send(fastify.swagger({}, opts.confIdx))
```

Issue #236, opened 2020-04-26
fastify-swagger multi-server option support


cyb3rs4pi3n:
## 🚀 Feature Proposal

Support for multi-server based on OpenAPI 3

## Motivation

Currently only host is supported. To test in multiple environments[local/dev/beta/prod] there is a need to specify multi-server support

## Example

```
servers: [
      {
        url: 'localhost',
        description: 'localhost'
      },
      {
        url: '{{dev-url}}',
        description: 'dev'
      }
    ]
```

reference: https://swagger.io/docs/specification/api-host-and-base-path

PR #237, opened 2020-04-26
add support for multi server

cyb3rs4pi3n:
Addresses https://github.com/fastify/fastify-swagger/issues/236

#### Checklist

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -19,6 +19,7 @@ module.exports = function (fastify, opts, next) {
 
   const info = opts.swagger.info || null
   const host = opts.swagger.host || null
+  const servers = opts.swagger.servers || null
   const schemes = opts.swagger.schemes || null
   const consumes = opts.swagger.consumes || null
   const produces = opts.swagger.produces || null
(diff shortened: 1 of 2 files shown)


mcollina:
This is missing docs and tests. I do not know which other features are missing between swagger and openapi, so I would be cautious.

dynamic.js
```diff
@@ -59,7 +60,7 @@ module.exports = function (fastify, opts, next) {
     // Base swagger info
     // this info is displayed in the swagger file
     // in the same order as here
-    swaggerObject.swagger = '2.0'
+    servers ? swaggerObject.openapi = '3.0.0' : swaggerObject.swagger = '2.0'
```
Eomm, line 63: OAS 3.0 is not supported by this plugin and so the `servers` syntax.

So this change would not be enough to support it

cyb3rs4pi3n: true. It worked for my use-case but as mcollina pointed out, this has to be thoroughly accounted for all the features that may be in use in the swagger 2.0 [currently looking into it]

Eomm: I was analysing how to solve the `$ref`, and I found this module [@apidevtools/swagger-parser](https://www.npmjs.com/package/@apidevtools/swagger-parser).

I think it could be useful to check the difference between the OAS2 and OAS3 formats.

Then, I thought what could be the macro-steps for this:

1. collect all the schemas
1. build one single schema (I working on it to support $ref and shared schemas)
1. translate a JSON schema _(draft 07 - since draft2019 is not broadly supported yet)_ to Swagger/OAS2.0 or OAS3 based on user options
1. validate the generated OAS object
1. render it

PR #238, opened 2020-05-05
Update fastify to v3

frikille:
* Updates fastify to v3
* Adds dependabot config
* Moves CI from Travis to Github Actions

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #161, opened 2019-03-09
wip: show file upload on swagger

Eomm:
This DRAFT PR is related to #47

It will support the possibility to set the `type=file` on the swagger UI

PR #239, opened 2020-05-06
$ref support - OAS 2.0 compliant

Eomm:
Opening this PR as POC:
when the user uses `$ref`, `swagger-ui` will do an HTTP GET to retrieve the external JSON.
So, to let this works, we need to:

- expose many routes as many `fastify.addSchema()` the user does
- the user should use `$id: hello` because if he uses `$id: http://example.com`.. the GET will go to http://example.com and it won't work!!
- write a good tutorial 😄 

I will rebase it to #238 
This plugin could benefit from the changes too: https://github.com/SkeLLLa/fastify-oas/issues/30

Writing this feature I will check the OAS 2.0 feature supported to add those who are missing like #161 

My personal achievements would be to fix all the "invalid schema error" that doesn't let me generate a [Postman collection](https://stackoverflow.com/questions/39072216/how-to-import-swagger-apis-into-postman) form the generated swagger


mcollina:
I think we should really try to resolve all the internal references if possible. I think it provides a better user experience, mainly because somebody can download all the files and store them in version control if they need to auto-generate a client.

Would it be opposed to that @Eomm? I think it's just a matter to resolve all the `$ref`

Eomm:
I would resolve URI like `$id: 'myawesome:schema'` and not URL `$id: 'http://my-server/schema'`

One single file output seems mandatory to me also to let users generate docs or other useful stuff

mcollina:
> I would resolve URI like $id: 'myawesome:schema' and not URL $id: 'http://my-server/schema'

Agreed.

> One single file output seems mandatory to me also to let users generate docs or other useful stuff

Agreed.

routes.js
```diff
@@ -29,6 +29,25 @@ function fastifySwagger (fastify, opts, next) {
     }
   })
 
+  // fastify.ready((err) => {
+  //   if (err) {
+  //     throw err
+  //   }
+  // })
+  const allSchemas = fastify.getSchemas()
```
SkeLLLa, line 37: @Eomm, do you know if in v3 this method was reworked, so it will return all schemes, even those what were registered in child scopes with fastify.register?
Cause in fastify v2 I had to go through all child scopes in order to get all of them like https://github.com/SkeLLLa/fastify-oas/blob/master/lib/openapi/index.js#L5

Eomm: nope, it is encapsulated and tested here:
https://github.com/fastify/fastify/blob/master/test/schema-feature.test.js#L505

I agree that we could think a nicer solution 👍

SkeLLLa: I think looping through children is quiet fine, unless you have schemes with same names. But in fastify-oas module there were no such issues, so it can be used.

Eomm: In think the onRegister hook could let us to avoid to use the symbol, I will give it a try

SkeLLLa: One more idea - is to add addSchema hook to Fastify. That will allow plugins to get schemes when they are registered. And, for example, instead of resolving them we could put schema to swagger `definitions` and use `$ref` inside swagger route to that definition.

dynamic.js
```diff
@@ -3,16 +3,30 @@
...
   const routes = []
+  const sharedSchemasMap = new Map()
+  const ref = Ref({ clone: true })
 
   fastify.addHook('onRoute', (routeOptions) => {
     routes.push(routeOptions)
   })
 
+  fastify.addHook('onRegister', async (instance) => {
+    await instance.ready() // wait the addSchema
```
mcollina, line 20: I'm a bit worried by this change. How would you expect this to work?

Eomm: onRegister is "fire and forget" and the await ready is blocked till the start

I should manage error

Eomm:
This pr is harder than I thought... now I understand why json-schema would like to standardize with OAS

mcollina:
> This pr is harder than I thought... now I understand why json-schema would like to standardize with OAS

Why?

dynamic.js
mcollina: Don't pass an `async` function here. Otherwise it would be completely unclear to the reader.

Eomm:
I wanted to replace `$ref` to point to local definitions, as suggested by the RFC.

But the swagger config support `$ref` to definitions only for the body; 

headers, params and query parameter do not support `$ref` out of the box so I must adopt another approach

Eomm:
Update: missing more tests, than it should be ready to review

dynamic.js
mcollina: isn't https://github.com/fastify/fastify/pull/2166 / `onReady` what would be best used?

Eomm: I thought it wasn't released yet, I will update 👍

Eomm:
I think this PR is ready for feedback

I would like a check by @SkeLLLa too if you could be so kind to give me your though

dynamic.js
```diff
@@ -3,16 +3,31 @@
...
   const routes = []
+  const sharedSchemasMap = new Map()
+  let ref
 
   fastify.addHook('onRoute', (routeOptions) => {
     routes.push(routeOptions)
   })
 
+  fastify.addHook('onRegister', async (instance) => {
+    instance.addHook('onReady', (done) => {
```
mcollina, line 20: Can you please add a comment on this?

dynamic.js
```diff
@@ -3,16 +3,31 @@
...
+      const allSchemas = instance.getSchemas()
+      for (const schemaId of Object.keys(allSchemas)) {
+        if (!sharedSchemasMap.has(schemaId)) {
```
mcollina, line 23: Can we detect if the id is being overridden by something else?

routes.js
```diff
@@ -38,6 +38,25 @@ function fastifySwagger (fastify, opts, next) {
     }
   })
 
+  // fastify.ready((err) => {
+  //   if (err) {
+  //     throw err
+  //   }
+  // })
```
mcollina, line 45: This should be removed.

test/json-schema.test.js
```diff
@@ -0,0 +1,111 @@
...
+          name: 'hello',
+          description: 'hello',
+          required: true,
+          type: 'file'
+        }])
+      })
+      .catch(function (err) {
+        t.fail(err)
+      })
+  })
```
mcollina, line 110: can you switch this test to async/await?

dynamic.js
SkeLLLa: Maybe here we can add instance prefix to schema name if it exists? In some cases I guess it could solve some schema conflicts.

Eomm: The problem here is that onRegister is receiving for all the register so in this case:

- register root - addSchema 1
  - child 1 registered - addSchema 2
     - subchild 2 registered - addSchema 3

For subchild 2 i will read schema 1, 2, 3
For child 1 schema 1 and 2
For root only schema 1

So it will be quite normal to have duplicate and I think it is ok since the `addSchema` already has a check to verify duplicate `$id`

SkeLLLa: I mean slightly different case.

* root - schema-1
  * child1 (prefix '/news') - item
  * child2 (prefix '/comments') - item

So if `item` schemas are registered in isolated scopes they will know now nothing about each other. However when you gather all them in shared schema map they will have the same id.

mcollina: Could you leave a TODO comment in here? Or open an issue? I think we might want/be able to have a way to get _all schemas defined at the current level_ so you do not get them all here.

Eomm: I agree and I will describe the case 👍

Eomm:
I must add some other commit.. I forgot the `response` 😱 
I'm working on it now

ahgentil:
I have my schemas in multiple files.
One of them has common stuff like:
```
  $id: 'http://example.com/',
  definitions: {
    uuid: {
      type: 'string',
      pattern: '^[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}$'
    }
...
```
Then inside other plugins' schemas I reference that:
```
id: { $ref: 'http://example.com/#/definitions/uuid' },
```

So, I just noticed that it blows up here:
https://github.com/fastify/fastify-swagger/blob/32d4f4295e1f40f64049b4951d04c19dd5d6097a/dynamic.js#L381

because, as the function name states `localRefResolve()`, this ends up returning `undefined`:
https://github.com/fastify/fastify-swagger/blob/32d4f4295e1f40f64049b4951d04c19dd5d6097a/dynamic.js#L392-L396

Should this $ref resolver be able to resolve any $ref?
The `uuid` definition is  in the `externalSchemas` param.

Eomm:
Could you open an issue with a complete example?
I get the point but it is time spending recreate it

ahgentil:
voilá https://github.com/fastify/fastify-swagger/issues/252

Issue #241, opened 2020-05-15
relative redirect needed, instead of an absolute one


PavelPolyakov:
## 🐛 Bug Report

Good day,

In our use case we serve services under multiple proxies.

Imagine we've a service `my-fav-service`, which serves some api AND also serves `swagger` under `my-fav-service/documentation/`. Later, we make a proxy to the service, which proxies all the requests from `https://example.com/public/api/my-fav-service` to `my-fav-service`. 
When we hit `https://example.com/public/api/my-fav-service/documentation` we are redirected to `https://example.com/documentation` because of [this](https://github.com/fastify/fastify-swagger/blob/8187bd074abfa90c30cec4240fffd611582b9240/routes.js#L18).

Which is clear - the redirect, currently, is absolute. I want to change this redirect to the relative, which will work fine, and submit PR.

So these tests should pass:
```js
t.only('/documentation should redirect to ./documentation/static/index.html', t => {
  t.plan(4)
  const fastify = Fastify()
  fastify.register(fastifySwagger, swaggerInfo)

  fastify.get('/', () => {})
  fastify.post('/', () => {})
  fastify.get('/example', opts1, () => {})
  fastify.post('/example', opts2, () => {})
  fastify.get('/parameters/:id', opts3, () => {})
  fastify.get('/example1', opts4, () => {})

  fastify.inject({
    method: 'GET',
    url: '/documentation'
  }, (err, res) => {
    t.error(err)
    t.strictEqual(res.statusCode, 302)
    t.strictEqual(res.headers.location, './documentation/static/index.html')
    t.is(typeof res.payload, 'string')
  })
})

t.only('/documentation/ should redirect to ../documentation/static/index.html', t => {
  t.plan(4)
  const fastify = Fastify()
  fastify.register(fastifySwagger, swaggerInfo)

  fastify.get('/', () => {})
  fastify.post('/', () => {})
  fastify.get('/example', opts1, () => {})
  fastify.post('/example', opts2, () => {})
  fastify.get('/parameters/:id', opts3, () => {})
  fastify.get('/example1', opts4, () => {})

  fastify.inject({
    method: 'GET',
    url: '/documentation'
  }, (err, res) => {
    t.error(err)
    t.strictEqual(res.statusCode, 302)
    t.strictEqual(res.headers.location, '../documentation/static/index.html')
    t.is(typeof res.payload, 'string')
  })
})
```

[...]

PR #247, opened 2020-05-16
make redirects from the swagger root relative

PavelPolyakov:
Resolves https://github.com/fastify/fastify-swagger/issues/241 .

Looking forward to your feedback.

#### Checklist

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -6,17 +6,26 @@ const fastifyStatic = require('fastify-static')
 // URI prefix to separate static assets for swagger UI
 const staticPrefix = '/static'
 
+function getRedirectPathForTheRootRoute (url) {
+  let redirectPath
+
+  if (url.substr(-1) === '/') {
+    redirectPath = `.${staticPrefix}/index.html`
+  } else {
+    const urlPathParts = url.split('/')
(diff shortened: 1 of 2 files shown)


PavelPolyakov:
I don't know why tests fail and how to rerun them. Tried on my local with `12.16.3` - everything is ok. Also tried with `v10`. Any idea?

PavelPolyakov:
k, after rebase I have these on my local, will fix

PavelPolyakov:
now it's good, first it was based on `req.raw.originalUrl` and it seems it disappeared in `fastify v3`, however `req.url` should be compatible with `v2` and `v3`

mcollina:
Thanks for figuring it out. We are preparing for the Fastify v3 release and we have landed that change in the meanwhile.

mcollina:
@Eomm can you take a look as well?

PavelPolyakov:
👋 @Eomm @mcollina would you be so kind to release another version of the `fastify-swagger@2` branch, which includes this fix?

`fastify v3` is not even released, and I can't use `fastify-swagger@3` with my correct `fastify v2` project because it requires `fastify v3`.

Eomm:
I will do later today after work 👍🏼

Eomm:
2.6.0 released in `2.x` branch

PR #256, opened 2020-06-14
Include swagger extensions in final swagger documentation

### LINKED ISSUE #249: Keep custom fields that start with 'x-' intact in swagger
## 🚀 Feature Proposal

I propose to include all fields that start with 'x-' (as per swagger spec https://swagger.io/docs/specification/2-0/swagger-extensions/) in the final swagger specification.  Currently, such extensions in parameters, responses, bodies, etc. are included (because they are passed verbatim), but the path object only keeps properties that are specifically hard-coded.

## Motivation

That would be useful for custom schema renderers that can utilize that additional information about paths.

## Example

```js
  fastify.route({
    method: 'GET',
    url: '/path',
    schema: {
      summary: 'Some method',
      response: {
        200: '...'
      },
      'x-custom-field': 'custom info'
    },
    ...
  })
```

This code should generate something like:

```json
{
  "paths": {
    "/path": {
      "get": {
        ...,
        "x-custom-field": "custom info"
      }
    }
  }
}
```

thomasthiebaud:
Fixes #249 

#### Checklist

- [ ] ~~documentation is changed or added~~

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -209,6 +209,12 @@ module.exports = function (fastify, opts, next) {
         if (schema.security) {
           swaggerMethod.security = schema.security
         }
+
+        Object.keys(schema).forEach(key => {
+          if (key.startsWith('x-')) {
+            swaggerMethod[key] = schema[key]
+          }
+        })
       }
 
       swaggerMethod.responses = genResponse(schema ? schema.response : null)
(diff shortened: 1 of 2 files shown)


dynamic.js
```diff
@@ -209,6 +209,12 @@ module.exports = function (fastify, opts, next) {
         if (schema.security) {
           swaggerMethod.security = schema.security
         }
+
+        Object.keys(schema).forEach(key => {
```
mcollina, line 213: would you mind to use a for...if loop instead?

thomasthiebaud: It should be fixed now

Issue #260, opened 2020-06-23
Type incorrectness


robcresswell:
## 🐛 Bug Report

The types listed for fastify-swagger are incorrect; TypeScript complains about the plugin being invalid when calling `register`, but when type checking is disabled it loads fine.

I can push a minimal repro after work, but wanted to open this for now

## To Reproduce

```ts
import Fastify from 'fastify'; // v3.0.0.rc-4
import * as fastifySwagger from 'fastify-swagger'; // v3.1.0

const fastify = Fastify();

fastify.register(fastifySwagger, {
  exposeRoute: true,
  swagger: {
    info: {
      title: process.env.npm_package_name ?? '',
      version: process.env.npm_package_version ?? '',
      description: process.env.npm_package_description,
    },
  },
});
```

## Expected behavior
No TS compilation errors

## Your Environment

- *node version*: 12
- *fastify version*: 3.0.0.rc-4
- *fastify-swagger*: 3.1.0
- *os*: Linux

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

robcresswell:
Absolutely! Will try and make some time ASAP

mcollina:
thanks!

robcresswell:
I've been looking into this, and I can't really find a pleasant way of doing it that is compatible with Fastify's export style. The types as they are now define a `default` export, which is _not_ representative of the actual export. For example, a default export in ES6 would be translated to

```
module.exports = {
  default: fastifySwagger //pseudo example
}
```

but in reality the export is

```
module.exports = fastifySwagger
```

This is complicated because as far as I can tell, top level exports in ES6 modules **must** be objects, they cannot be callables. Unfortunately, digging into the issue, I realised that every repo in fastify-land seems to have this pattern.

In TypeScript this issue is solvable using ```esModuleInterop: true```, which causes TypeScript to generate a bunch of synthetic imports, but it feels very weird to me that the design of a library should enforce the project using it to be configured in a specific way.

robcresswell:
I could start overloading the exports, but those get weird / ugly very fast, and it feels like using JS's dynamic powers for evil.

PR #264, opened 2020-07-08
fix: Correct types

robcresswell:
Partial revert of https://github.com/fastify/fastify-swagger/pull/238/commits/530514a394c58fa8d26b5084a813de2b49a42d17 which broke the types declared in this plugin. I can't find anything in the PR or commit message that explains why this was done, but this module does not have a default export.

Fixes: #260

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -62,4 +62,4 @@ type SwaggerOptions = (FastifyStaticSwaggerOptions | FastifyDynamicSwaggerOption
 
 declare const fastifySwagger: FastifyPlugin<SwaggerOptions>
 
-export default fastifySwagger;
+export = fastifySwagger;
--- a/test/types/http2-types.test.ts
+++ b/test/types/http2-types.test.ts
@@ -1,5 +1,5 @@
 import fastify from 'fastify';
-import fastifySwagger from '../..';
+import fastifySwagger = require('../..');
 
 const app = fastify({
   http2: true
--- a/test/types/types.test.ts
+++ b/test/types/types.test.ts
@@ -1,5 +1,5 @@
 import fastify from 'fastify';
-import fastifySwagger from '../..';
+import fastifySwagger = require('../..');
 
 const app = fastify();
 


robcresswell:
I think I'll need to rewrite the type definitions, as I can't figure out a good way to do the top level namespace export, and also export the SwaggerOptions type, to test it.

mcollina:
is this ok to land?

cc @Ethan-Arrowood ?

Ethan-Arrowood:
If the module itself doesn't have a default export then we just need to `export fastifySwagger` and you'd import it like `import { fastifySwagger } from 'fastify-swagger'`. I don't like seeing the `import require` syntax so i think there is a better solution here.

robcresswell:
Agreed, it isn't a good solution; but its not a lie, at least :smile: 

Frankly, with ESM, named exports everywhere makes life easier, but its a breaking change for the whole ecosystem, which isn't ideal considering 3.x _just happened_.

Ethan-Arrowood:
Correct me if I'm wrong but how does this plugin differ from others? It is exporting `fp` as expected: https://github.com/fastify/fastify-swagger/blob/074204df83bb29ce368aba37e294d84f892fc6e4/index.js#L25

I think the only change that needs to happen is 
`export default` to `export =`

robcresswell:
Because `import fastifySwagger from '../..';` is a default export, which is not whats provided if I change the types to `export =`.

It seemed like the most agreeable route was to convert to what the codebase had previously

Ethan-Arrowood:
What is `export =` if it is not a default export?

robcresswell:
`export =` is a TypeScript workaround for ESM / CJS module compatibility, as is `import = require()`. `export =` isn't part of the ESM spec; this is what I'm getting at when I'm trying to explain that the whole `module.exports =` as a callable isn't interoperable, and ideally should be dropped, to the best of my knowledge.

A default export (`export default 'foo'`) is effectively the same as calling `module.exports = { default: 'foo' }`. If you look at ESM / CJS code thats been transformed for compatibility, that's usually how its implemented.

Like I said in my earlier comment, its actually a problem with all of Fastify; the docs have been written assuming that the person consuming fastify has `esModuleInterop: true` in their tsconfig, but I'm not sure if thats intentional? For example, if you look at the main Typescript docs in the Fastify repo, there are examples like 

```ts
import fs from 'fs';
```

but that indicates a default export, which `fs` does not have. The correct import _should_ be

```ts
import * as fs from 'fs';
```

The other option is to make it explicit in the docs that Fastify expects that config option to be set, I just personally feel that libraries should not enforce that on consumers, because it has implications on other aspects of the codebase, like how you stub module boundaries in tests.

robcresswell:
Hey @Ethan-Arrowood, any more thoughts on this? No pressure if you're busy, just making sure its vaguely on your radar.

Ethan-Arrowood:
@robcresswell great explanation. thank you for this! I'm definitely still learning myself so this is useful information. Unfortunately I would have to say that it is recommended that users use the `esModuleInterop: true` for fastify. I always have it turned on myself and have never had a reason to have it off (in fact I think even the config generated by typescript has it enabled by default). If you feel strongly about this and want to see how we can improve fastify as well as fastify plugins please feel free to continue this conversation. We are avoiding breaking changes at this moment but it never hurts to track discussions like this for the future of Fastify and its ecosystem!

fox1t:
Hi, sorry for writing on a closed PR, but I found a reference for it in another issue. 
@robcresswell may I ask you why you didn't used `import * as fastifySwagger  from ('../..');`?

It is the correct syntax to deal with CJS modules today: it imports a namespace and what FP returns is a namespace.

robcresswell:
Hey @fox1t,

Just for the reasons in the commit message; I focused on reverting rather than changing.

I'm new to this community / project, so it seemed the best way to avoid ruffling feathers.

fox1t:
Ok! Thanks for you answer! Just to make clear `import * as smth from 'smth'` and `import smth = require('smth')` are basically the same, but the latter is the older syntax that is (almost) not used anymore. ^^

robcresswell:
@fox1t They are not the same. They are _similar_, but `import * as` on a callable is invalid under the ES spec, and will throw an error in TS if you have esmoduleinterop disabled.

fox1t:
If the typings were correct, they would be the same: and that's the point I am trying to make here. TS v 2.7 introduced that additional check when esModuleInterop is disabled to respect ESM approach that imposes that namespaces are never callable.

robcresswell:
Perhaps I am misreading then? From what I can see https://github.com/fastify/fastify-plugin/blob/master/plugin.js seems to return a callable as its export, as does https://github.com/fastify/fastify-swagger/blob/master/index.js

fox1t:
No, you are right and your fix is absolutely fine. I am writing a document to recap all the things we have to pay attention on and hopefully to not make more mistakes in future.

robcresswell:
Ah cool, that sounds interesting! If some better ideas than mine come up, I'm happy to help push them.

PR #266, opened 2020-07-10
Copy response description from schema

wandersonwhcr:
Fastify defines response description using `route.schema.response.${code}.description` and Swagger uses `route.responses.${code}.description`.

This PR copies `description` property from `schema` to "response code" object and removes it from `schema`. This works only if the response has a `type` or `$ref` property, avoiding to *always* fill with "Default Response".

https://swagger.io/docs/specification/describing-responses/#body

#### Checklist

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -285,6 +285,10 @@ module.exports = function (fastify, opts, next) {
           responsesContainer[key] = {
             schema: resolved
           }
+          if (resolved.description) {
+            responsesContainer[key].description = resolved.description
+            delete resolved.description
+          }
         } else {
           responsesContainer[key] = resolved
         }
--- a/test/swagger.js
+++ b/test/swagger.js
@@ -441,7 +441,7 @@ test('deprecated route', t => {
 })
 
 test('route meta info', t => {
-  t.plan(8)
+  t.plan(9)
   const fastify = Fastify()
 
   fastify.register(fastifySwagger, swaggerInfo)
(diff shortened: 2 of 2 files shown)


wandersonwhcr:
There's no need to change docs ;)

dynamic.js
```diff
@@ -285,6 +285,10 @@ module.exports = function (fastify, opts, next) {
           responsesContainer[key] = {
             schema: resolved
           }
+          if (resolved.description) {
+            responsesContainer[key].description = resolved.description
+            delete resolved.description
```
Eomm, line 290: I fear that this will change the schema of the user, we could just avoid to delete this field and the swagger output should just work as now

mcollina:
@Eomm I prefer for the fixes to go to master first and eventually be backported, not vice-versa.

Eomm:
The code should be quite different cause the $ref support, so I doubt a cherry pick will be doable. 
I think a new pr will be needed in any case

mcollina:
Anyway, this needs fixing for master first.

dynamic.js
wandersonwhcr: @Eomm so... can I remove the line 290?

Eomm: Could you test it?

If it is true, that the json object of the user is changed, I would remove the line since it won't affect the swagger API interface

Eomm:
Superseded by #301

Issue #258, opened 2020-06-16
100% code coverage


mcollina:
As titled!

iandrc:
Hey @mcollina 

I would like to try this one if you don't mind. Are there any pitfalls or something else I should know about?

mcollina:
Not that I know of! Go for it!

iandrc:
@mcollina 

See [#267](https://github.com/fastify/fastify-swagger/pull/267)

I've done a bit of a job, though I need a review and, probably, some help from you or another team member

zekth:
Reopening since the coverage has dropped with: https://github.com/fastify/fastify-swagger/pull/333
and the ci has coveralls landed in : https://github.com/fastify/fastify-swagger/pull/348/files

climba03003:
Closing as it is bug from the `script`.
```Shell
tap --100 test/**/*.js
```

Missing quote for `**` will use the shell evaluation of files, and some of the system may not handle `globstar` as default (e.g. coveralls). As a result, the coverage result drops.
Fixed in #351

PR #267, opened 2020-07-12
increase code coverage to 100%

iandrc:
Currently, 4 out of 4 files have 100% coverage:

| File       | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line # |
|------------|---------|----------|---------|---------|------------------|
| All files  |   100 |    100 |     100 |   100 |                  |
| dynamic.js |   100 |    100 |     100 |   100 |               |
|   index.js |     100 |      100 |     100 |     100 |                  |
|  routes.js |     100 |      100 |     100 |     100 |                  |
|  static.js |     100 |      100 |     100 |     100 |                  |

Fixes: https://github.com/fastify/fastify-swagger/issues/258

#### Checklist


static.js
```diff
@@ -65,6 +65,8 @@ module.exports = function (fastify, opts, next) {
     }
 
     fastify.register(require('./routes'), options)
+  } else {
+    console.log()
```
mcollina, line 69: why this empty `console.log` was added here?

test/static.js
```diff
@@ -43,17 +43,88 @@ test('specification validation check works', t => {
...
+    exposeRoute: true,
+    mode: 'static',
+    specification: {
+      path: './examples/example-static-specification.yaml',
+      baseDir: 1
+    }
+  }
+
+  t.plan(1)
+  const fastify = new Fastify()
```
mcollina, line 90: remove `new`

test/static.js
```diff
@@ -304,3 +381,242 @@ test('/documentation/:file should be served from custom location with trailing s
...
+    },
+    (err, res) => {
+      t.error(err)
+      t.is(typeof res.payload, 'string')
+      t.is(res.headers['content-type'], 'application/json; charset=utf-8')
+      try {
+        t.pass('valid swagger json')
+      } catch (err) {
+        t.fail(err)
+      }
```
mcollina, line 478: this try-catch is superflous

test/static.js
```diff
@@ -304,3 +381,242 @@ test('/documentation/:file should be served from custom location with trailing s
...
```
mcollina, line 461: this try-catch is superflous

test/static.js
```diff
@@ -304,3 +381,242 @@ test('/documentation/:file should be served from custom location with trailing s
...
+    (err, res) => {
+      t.error(err)
+      t.is(typeof res.payload, 'string')
+      t.is(res.headers['content-type'], 'application/x-yaml')
+      try {
+        yaml.safeLoad(res.payload)
+        t.pass('valid swagger yaml')
+      } catch (err) {
+        t.fail(err)
+      }
```
mcollina, line 430: this try-catch is superflous

test/static.js
```diff
@@ -304,3 +381,242 @@ test('/documentation/:file should be served from custom location with trailing s
...
+    },
+    (err, res) => {
+      t.error(err)
+      t.is(typeof res.payload, 'string')
+      t.is(res.headers['content-type'], 'application/json; charset=utf-8')
+      try {
+        t.pass('valid swagger json')
+      } catch (err) {
+        t.fail(err)
+      }
```
mcollina, line 557: this try-catch is superflous

test/static.js
```diff
@@ -304,3 +381,242 @@ test('/documentation/:file should be served from custom location with trailing s
...
```
mcollina, line 574: this try-catch is superflous

routes.js
```diff
@@ -68,7 +68,7 @@ function fastifySwagger (fastify, opts, next) {
     schema: { hide: true },
     handler: function (req, reply) {
       const file = req.params['*']
-      if (file === '') {
+      if (file === 'null') {
```
mcollina, line 71: why this change? I think this was correct before

test/static.js
```diff
@@ -304,3 +381,242 @@ test('/documentation/:file should be served from custom location with trailing s
...
+    }
+  )
+})
+
+test('inserts default opts in fastifySwagger', t => {
+  t.plan(1)
+  const fastify = Fastify()
+  const next = () => console.log()
```
mcollina, line 614: why?

test/static.js
```diff
@@ -304,3 +381,242 @@ test('/documentation/:file should be served from custom location with trailing s
...
+
+  // eslint-disable-next-line
+  fastify.register(Function(fastifySwagger(fastify, null, next)))
```
mcollina, line 617: Why the `Function`?

mcollina:
> ** I have some issues with dynamic.js would be nice if you can help me out with it.**

what issues?

static.js
iandrc: Otherwise, the coverage tool says that `else path not taken` and treats it as an uncovered branch. I didn't know how to do it the other way. That's why I inserted an empty `console.log`. Though, that's not the best option. 

I saw some tests in `fastify` repo, you don't have such a problem. Do you ignore such messages from the coverage somehow?

routes.js
iandrc: That's why I asked you to take a look. Actually, otherwise, I can't reach the code inside the `if` statement. I believe that's because it uses the default redirection rules defined above.

How do you suggest to change this line so it a) works correctly, b) does not harm the coverage?

test/static.js
iandrc: I needed to pass `next` function as a param to `fastifySwagger`, and the `next` returns `void`. I didn't find a better solution to this.

test/static.js
iandrc: If we don't use a `Function` constructor, `avvio` throws an error saying that:
```
 FAIL  test/static.js
✖ plugin must be a function or a promise
node_modules/avvio/boot.js                                                                       
  192 | function assertPlugin (plugin) {                                                           
  193 |   if (!(plugin && (typeof plugin === 'function' || typeof plugin.then === 'function'))) {  
> 194 |     throw new Error('plugin must be a function or a promise')                              
      | ----------^                                                                                
  195 |   }                                                                                        
  196 | }                                                                                          
  197 | 
```

So, if we remove the `Function` for this test, it'll fail

iandrc:
> > ** I have some issues with dynamic.js would be nice if you can help me out with it.**
> 
> what issues?

Yeah, I've spent a couple of days trying to cover all the remaining lines/statements/branches inside `dynamic.js` file but I just can't reach them. So, it would be nice to have some new ideas from you or someone else how I can do it. 
Additionally, there are `8` uncovered branches (approx. half of them do not have an `else` block). What should I do with that half? Add some kind of `else { console.log() }` or we can ignore them?

mcollina:
Which lines are unreacheable? You can paste links to the line in the issue.

---

> Additionally, there are 8 uncovered branches (approx. half of them do not have an else block). What should I do with that half? Add some kind of else { console.log() }

I don't understand why adding an `else` and a `console.log` helps. You need to add a test where the main condition in the `if` is not triggered.

iandrc:
> Which lines are unreacheable? You can paste links to the line in the issue.
> 
> > Additionally, there are 8 uncovered branches (approx. half of them do not have an else block). What should I do with that half? Add some kind of else { console.log() }
> 
> I don't understand why adding an `else` and a `console.log` helps. You need to add a test where the main condition in the `if` is not triggered.

I'll try more then

routes.js
mcollina: I think you just need to send your http request to `/documentation` or `/documentation` so that the `*` param would be empty.

test/static.js
mcollina: _never_ use `console.log`. You can add an assertion there with `t.error(err)` to assert that `next` is called without errors.

routes.js
iandrc: Nope, it should be smth like: `/documentation/a` at least in order to trigger those lines. Therefore, they won't be empty. 
That's why I've changed that check to 
```
if (file === 'null')
```

I tried different approaches but still couldn't cover those lines 'as is'. So, if you have any ideas just let me know

dynamic.js
```diff
@@ -218,6 +218,7 @@ module.exports = function (fastify, opts, next) {
       }
 
       swaggerMethod.responses = genResponse(schema ? schema.response : null)
+      /* istanbul ignore next */
```
iandrc, line 221: I've ignored the following check 'cause we'll always have a `swaggerRoute` in place, therefore, we have no need to take `else` path

dynamic.js
```diff
@@ -281,14 +282,17 @@ module.exports = function (fastify, opts, next) {
         const rawJsonSchema = fastifyResponseJson[key]
         const resolved = ref.resolve(rawJsonSchema)
 
+        /* istanbul ignore next */
```
iandrc, line 285: The following `if-else` is ignored as a `type` is somehow inserted under the hood. Therefore, we'll never fall into an `else` as far as I understood

dynamic.js
```diff
@@ -281,14 +282,17 @@ module.exports = function (fastify, opts, next) {
...
         if (resolved.type || resolved.$ref) {
           responsesContainer[key] = {
             schema: resolved
           }
         } else {
+          /* istanbul ignore next */
           responsesContainer[key] = resolved
         }
 
+        /* istanbul ignore next */
```
iandrc, line 295: We'll never have a `description` unless we directly specify it before the `if`

routes.js
mcollina: What is called when you call `/documentation` or `/documentation/` then?

iandrc: Code on line `23` if I understand correctly:
```
fastify.route({
    url: '/',
    method: 'GET',
    schema: { hide: true },
    handler: (req, reply) => {
      reply.redirect(getRedirectPathForTheRootRoute(req.raw.url))
    }
  })
```

dynamic.js
mcollina: Then let's drop the if?

dynamic.js
mcollina: Let's drop the if.

dynamic.js
mcollina: let's drop the if.

routes.js
mcollina: Then we can drop the if.

dynamic.js
iandrc: I'll do so

dynamic.js
iandrc: Ok then

mcollina:
how are we for code coverage now?

iandrc:
Hey @mcollina 

We're approx. at the same point. I'm currently trying to cover the remaining statements and branches :)
Hopefully, I'll push them today or maybe tomorrow

dynamic.js
```diff
@@ -106,6 +106,7 @@ module.exports = function (fastify, opts, next) {
       swaggerObject.externalDocs = externalDocs
     }
 
+    /* istanbul ignore next */
```
mcollina, line 109: Why is this needed?

dynamic.js
```diff
@@ -281,17 +281,11 @@ module.exports = function (fastify, opts, next) {
...
-        } else {
-          responsesContainer[key] = resolved
+        responsesContainer[key] = {
+          schema: resolved
         }
 
-        if (!responsesContainer[key].description) {
-          responsesContainer[key].description = 'Default Response'
-        }
+        responsesContainer[key].description = 'Default Response'
```
mcollina, line 288: merge this with the object definition at line 285.

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "prepare": "node prepare-swagger-ui",
-    "test": "standard && tap test/*.js && npm run typescript",
+    "test": "standard && tap --100 test/*.js && npm run typescript",
```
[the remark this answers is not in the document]
mcollina: Perfect!

dynamic.js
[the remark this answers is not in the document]
mcollina: You should add a test where `ref` is falsy.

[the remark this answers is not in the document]
iandrc: @mcollina 
The `ref` is inside `dynamic.js` set on [line 13](https://github.com/fastify/fastify-swagger/blob/master/dynamic.js#L13) and has `undefined` value. 

I believe that was done on purpose. If we initialize it with some values, it may affect the existing logic. Therefore, we cannot change its value. 

So, what do you suggest?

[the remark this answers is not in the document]
mcollina: I mean there should be a test where we are calling the `app.swagger()` function twice.

dynamic.js
```diff
@@ -106,22 +106,20 @@ module.exports = function (fastify, opts, next) {
       swaggerObject.externalDocs = externalDocs
     }
 
-    if (!ref) {
```
iandrc, line 109: I've played with code a bit more and found out that we don't need this `if` clause. The `swagger` function is called only once at the very beginning, and the second call does not allow us to trigger it once again. Therefore, we're quite safe to remove it, it should not break any logic.

iandrc:
Hey @mcollina 

Should I add something or this can be merged to the main branch? Do I need to request some more reviews?

mcollina:
I was waiting for more reviews.. but not really needed! Landed!

PR #270, opened 2020-07-16
Adds better support for TS and ES modules imports

fox1t:
Hi, I am opening this PR in order to show what has been discussed here: https://github.com/fastify/fastify/issues/2403#issuecomment-659311897
Here we support all of the cases that are show in that issue. 
For the TS side:
1. `import * as fastifySwagger from 'fastify-swagger'` 
2. `import fastifySwagger from 'fastify-swagger'`: this works thanks to adding `.default` export to `module.exports` on JS side.
3. the "deprecated" `import fastifySwagger= require('fastify-swagger)`

On the JS side it adds `.default` property to the `module.exports` in order to support also ES default imports.

We can use this PR to further check if this can fit for Fastify's ecosystem and find potential bugs and quirks.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,11 @@ function fastifySwagger (fastify, opts, next) {
   }
 }
 
-module.exports = fp(fastifySwagger, {
+const plugin = fp(fastifySwagger, {
   fastify: '>=3.x',
   name: 'fastify-swagger'
 })
+
+plugin.default = fastifySwagger
+
+module.exports = plugin
--- a/index.d.ts
+++ b/index.d.ts
@@ -60,6 +60,8 @@ declare module 'fastify' {
 
 type SwaggerOptions = (FastifyStaticSwaggerOptions | FastifyDynamicSwaggerOptions)
 
-declare const fastifySwagger: FastifyPlugin<SwaggerOptions>
+declare const fastifySwagger: FastifyPlugin<SwaggerOptions> & {
+  default: FastifyPlugin<SwaggerOptions>;
+};
 
 export = fastifySwagger;
--- a/test/default-export.js
+++ b/test/default-export.js
@@ -1,0 +1,55 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('fastify')
+const Swagger = require('swagger-parser')
+const fastifySwagger = require('../index')
+
+const swaggerInfo = {
+  swagger: {
+    info: {
+      title: 'Test swagger',
+      description: 'testing the fastify swagger api',
+      version: '0.1.0'
+    },
+    host: 'localhost',
+    schemes: ['http'],
+    consumes: ['application/json'],
+    produces: ['application/json'],
+    securityDefinitions: {
+      apiKey: {
+        type: 'apiKey',
+        name: 'apiKey',
+        in: 'header'
+      }
+    }
+  },
+  exposeRoute: true
+}
+
+test('/documentation/json route default export', t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  fastify.register(fastifySwagger.default, swaggerInfo)
+
+  fastify.get('/', () => {})
+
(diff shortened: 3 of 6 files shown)


index.js
```diff
@@ -22,7 +22,11 @@ function fastifySwagger (fastify, opts, next) {
   }
 }
 
-module.exports = fp(fastifySwagger, {
+const plugin = fp(fastifySwagger, {
   fastify: '>=3.x',
   name: 'fastify-swagger'
 })
+
+plugin.default = fastifySwagger
```
mcollina, line 30: I think we should do this within `fastify-plugin` so all fastify plugins automatically work instead.

fox1t: Yes, I agree with you. I am going to make a PR on `fastify-plugin` later since I am still figuring out the correct types that do not break anything. We will also need a PR to Fastify itself, for the `.regsiter()` method typings. Roughly something like this:
```
import { FastifyInstance } from './instance'
import { RawServerBase, RawRequestDefaultExpression, RawReplyDefaultExpression, RawServerDefault } from './utils'

/**
 * FastifyPluginCallback
 *
 * Fastify allows the user to extend its functionalities with plugins. A plugin can be a set of routes, a server decorator or whatever. To activate plugins, use the `fastify.register()` method.
 */

interface FastifyPluginCallback<Options extends FastifyPluginOptions = {}, Server extends RawServerBase = RawServerDefault> {
  (
    instance: FastifyInstance<Server, RawRequestDefaultExpression<Server>, RawReplyDefaultExpression<Server>>,
    opts: Options,
    next: (err?: Error) => void
  ): void
  default: (
    instance: FastifyInstance<Server, RawRequestDefaultExpression<Server>, RawReplyDefaultExpression<Server>>,
    opts: Options,
    next: (err?: Error) => void
  ) => void
}

/**
 * FastifyPluginAsync
 *
 * Fastify allows the user to extend its functionalities with plugins. A plugin can be a set of routes, a server decorator or whatever. To activate plugins, use the `fastify.register()` method.
 */
interface FastifyPluginAsync<Options extends FastifyPluginOptions = {}, Server extends RawServerBase = RawServerDefault> {
  (
    instance: FastifyInstance<Server, RawRequestDefaultExpression<Server>, RawReplyDefaultExpression<Server>>,
    opts: Options
  ): Promise<void>
  default:  (
    instance: FastifyInstance<Server, RawRequestDefaultExpression<Server>, RawReplyDefaultExpression<Server>>,
    opts: Options
  ) => Promise<void>
}

/**
 * Generic plugin type.
 * @deprecated union type doesn't work well with type inference in TS and is therefore deprecated in favor of explicit types. See FastifyRegister.
 */
export type FastifyPlugin<Options extends FastifyPluginOptions = {}> = FastifyPluginCallback<Options> | FastifyPluginAsync<Options>

export interface FastifyPluginOptions {
  [key: string]: any;
}

```

fox1t: However I think that we have to encourage all plugin developers to always export also the default prop in order to be sure that even if FP is not used, the plugin will work in every scenario.
What do you think about this? Do you think it is ok to put a note in the docs (not only in the TS section but also in the JS one)?

mcollina: Let's do the changes to fastify itself, to fastify-plugin, and the note docs.

fox1t: Ok just please check the quirks at https://github.com/fastify/fastify/issues/2403#issuecomment-659470548

I really want to make sure that we all agree and are on the same page, since I finally got the big picture. :)

harveyconnor:
LGBT

index.js
mcollina: any update @fox1t?

fox1t: Hi, I just published a document that recaps TS/ES modules. We can use it to decide what we want to achieve.

fox1t:
Closing this in favor of [282](https://github.com/fastify/fastify-swagger/pull/282)

PR #276, opened 2020-07-28
Export ts types

hendrysadrak:
Make FastifyDynamicSwaggerOptions, FastifyStaticSwaggerOptions & SwaggerOptions exported for cases like:

```ts
import fastifySwagger, { SwaggerOptions } from 'fastify-swagger'

const options: SwaggerOptions = {
  ...
}

app.register(fastifySwagger, options)
```

 This was possible in v2

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -15,7 +15,7 @@ interface FastifySwaggerOptions {
   exposeRoute?: boolean;
 }
 
-interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
+export interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
   mode?: 'dynamic';
   swagger?: Partial<SwaggerSchema.Spec>;
   /**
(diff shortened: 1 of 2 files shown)


mcollina:
This fails our typescritp tests:
```
   ✖  65:0  An export assignment cannot be used in a module with other exported elements.  
```

You can run them with `npm test`. Also, can you add a test for this feature?

sorakthunly:
This is very much needed.

I have to now do this needlessly when the types can just be exported

```ts
import swagger from 'fastify-swagger';

export const swaggerOptions: Parameters<typeof swagger>[1] = { ... }
```

hendrysadrak:
Should i add tests for FastifyDynamicSwaggerOptions, FastifyStaticSwaggerOptions, StaticPathSpec, StaticDocumentSpec too?

mcollina:
That would be nice! Are you adding them to this PR or another?

hendrysadrak:
Because this PR already solves the initial problem, I can do it in a followup PR

PR #282, opened 2020-08-10
Adds the "infamous triplet" and exports default and names exports in TypeScript types

fox1t:
This PR is the result of the ongoing talks about supporting all import contexts for the fastify ecosystem. We can use it as a case study for other core plugins.

What is important to note is that it is not enough to add this code only at JS, since the exported typings need to be updated too.

- Adds the "infamous triplet" in `module.exports`
- Replaces namespace export with "infamous triplet" in TypeScript typings
- Adds test for all TS imports
- Adds test for ESM context

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,11 @@ function fastifySwagger (fastify, opts, next) {
   }
 }
 
-module.exports = fp(fastifySwagger, {
+const fastifySwaggerPlugin = fp(fastifySwagger, {
   fastify: '>=3.x',
   name: 'fastify-swagger'
 })
+
+module.exports = fastifySwaggerPlugin
+module.exports.default = fastifySwaggerPlugin
+module.exports.fastifySwagger = fastifySwaggerPlugin
--- a/index.js
+++ b/index.js
@@ -22,11 +22,7 @@ function fastifySwagger (fastify, opts, next) {
   }
 }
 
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "prepare": "node prepare-swagger-ui",
-    "test": "standard && tap --100 test/*.js && npm run typescript",
+    "test": "standard && tap --no-esm --100 test/*.js test/esm/*.js && npm run typescript",
```
mcollina, line 8: https://github.com/fastify/fastify/blob/master/.taprc < use this instead.

fox1t: Done

index.js
```diff
@@ -22,7 +22,11 @@ function fastifySwagger (fastify, opts, next) {
...
 
-module.exports = fp(fastifySwagger, {
+const fastifySwaggerPlugin = fp(fastifySwagger, {
   fastify: '>=3.x',
   name: 'fastify-swagger'
 })
+
+module.exports = fastifySwaggerPlugin
+module.exports.default = fastifySwaggerPlugin
+module.exports.fastifySwagger = fastifySwaggerPlugin
```
mcollina, line 32: I don't think these last two are needed with the new fastify-plugin release

fox1t: I confirm that we don't need it. Removed.

PR #289, opened 2020-08-22
use the response schema's custom description if present

dound:
Previously, the response schema's description was ignored (though it was present in fastify-swagger examples, it was not shown on the generated documentation webpage). Now, instead of always showing "Default Response" we show the value of the `description` field, if provided.


climba03003:
this PR should be closed as it duplicate with #301

Eomm:
Superseded by #301

Issue #299, opened 2020-09-21
Response description prop is not appended correctly to schema definition


MetCoder95:
## 🐛 Bug Report

When defining a response schema definition,
and compiling to swagger spec, the `description` prop is not appended correctly ending with unexpected behavior where the Swagger UI doesn't show the custom description provided

## To Reproduce

**Input**:
```js
{
   "response":{
      200:{
         "type":"object",
         "properties":{
            "hello":{
               "type":"string"
            }
         },
         "description":"some description"
      }
   }
}
```

**Output**:
```json
{
   "responses":{
      "201":{
         "schema":{
            "type":"object",
            "properties":{
               "hello":{
                  "type":"string"
               }
            },
            "description":"some description"
         },
         "description":"Default Response"
      }
   }
}
```

## Expected behavior

**Input**:
```js
{
   "response":{
      "200":{
         "type":"object",
         "properties":{
            "hello":{
               "type":"string"
            }
         },
         "description":"some description"
      }
   }
}
```

**Output**:
```json
{
   "responses":{
      "201":{
         "schema":{
            "type":"object",
            "properties":{
               "hello":{
                  "type":"string"
               }
            }
         },
         "description":"some description"
      }
   }
}
```

## Your Environment

- *node version*: 14.11
- *fastify version*: 3.4.1
- *fastify-swagger version*: 3.3.0
- *os*: Mac
- *any other relevant information*

zekth:
Looks like this issue come from those lines : https://github.com/fastify/fastify-swagger/blob/3fb41c0ea2cc44ec917e9abfc80525e7412c30ed/dynamic.js#L168-L170

@Eomm is there a particular reason to do this?

Eomm:
The `description` is [mandatory](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#responseObject) and it must be at `2xx` level.

Valid:
```
      responses:
        "400":
          description: "Invalid ID supplied"
        "404":
          description: "Pet not found"
        "405":
          description: "Validation exception"
```

Invalid:

```
      responses:
        description: "Invalid supplied"
```

So this desired output would break swagger-ui I think:

```json
"responses": {
	"201": {
		"schema": {
			"type": "object",
			"properties": {
				"hello": {
					"type": "string"
				}
         	}
    	},
    	"description": "some description"
	}
}
```

MetCoder95:
But that output is fin, isn't it?

I mean, when I have this output:
```json
[...]

Issue #250, opened 2020-05-24
Manage 2xx in response


Eomm:
## 🐛 Bug Report

The [swagger spec](https://swagger.io/docs/specification/2-0/describing-responses/) doesn't support `2xx` format code and `fastify.swagger()` doesn't manage them.

Right now the UI doesn't break, it works fine: 

![image](https://user-images.githubusercontent.com/11404065/82751521-16c87280-9db8-11ea-9393-c3e7f08b70f4.png)

But the yaml is not valid one:

from http://localhost:3000/foobar/yaml to https://editor.swagger.io/

![image](https://user-images.githubusercontent.com/11404065/82751544-46777a80-9db8-11ea-9751-d438ac0d5910.png)

A valid schema is needed to provide the yaml output to other tools and automate some API building (like collection in Postman)

## To Reproduce

```js
const fastify = Fastify({ logger: true })

fastify.register(fastifySwagger, {
  exposeRoute: true,
  routePrefix: '/foobar'
})

fastify.get('/', {
  schema: {
    response: {
      '2xx': {
        hello: { type: 'string' }
      }
    }
  }
},
async () => 'hello')

fastify.listen(3000)
```

## Expected behavior

Produce a valid swagger output

I think this could be a flag that `forceCoertion` to apply some transformation to have a valid output.

Eomm:
I think the algorithm I would implement would be:

if there is a `2xx` it will be transformed to `200` family,
if there is a `4xx` it will be transformed to `400` family and so on... 

The issue could be if the user defines:
```
2xx: schema1
200: schema2
```

In this case, the code should throw an error at startup and warn the user. - In this use case, the dev is still in dev environment.

The user to fix this case should set up a new `defaultXXFamily` that we could receive from the plugin config

PR #301, opened 2020-09-25
Fix bug #299 Response description prop is not appended correctly to schema definition

maxidr:
Added support for description in response keys (status code) 

Like this example (present in bug #299):

```javascript
response: {
  200: {
    type: 'object',
    properties: {
      hello: { type: 'string' },
    },
    description: 'some description'
  }
}
```

### DIFF
--- a/test/json-schema.test.js
+++ b/test/json-schema.test.js
@@ -106,3 +106,58 @@ test('support file in json schema', async t => {
     type: 'file'
   }])
 })
+
+test('support response description', async t => {
+  const opts8 = {
+    schema: {
+      response: {
+        200: {
+          description: 'Response OK!',
+          type: 'object'
+        }
+      }
+    }
+  }
+
+  const fastify = Fastify()
+  fastify.register(fastifySwagger, {
+    routePrefix: '/docs',
+    exposeRoute: true
+  })
+  fastify.get('/', opts8, () => {})
+
+  await fastify.ready()
+
+  const swaggerObject = fastify.swagger()
+  const api = await Swagger.validate(swaggerObject)
+
+  const definedPath = api.paths['/'].get
+  t.same(definedPath.responses['200'].description, 'Response OK!')
+})
+
+test('response default description', async t => {
+  const opts9 = {
+    schema: {
+      response: {
+        200: {
+          type: 'object'
+        }
+      }
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


test/json-schema.test.js
```diff
@@ -106,3 +106,58 @@ test('support file in json schema', async t => {
     type: 'file'
   }])
 })
+
+test('support response description', async t => {
+  const opts8 = {
+    schema: {
+      response: {
+        200: {
```
Eomm, line 114: Could you add a test with `'2xx'`?
That pattern is used a lot and there is a conversion to manage it

maxidr: Hi @Eomm 

When tried to use '2xx' in the response, the swagger validator fails with:

```
Swagger schema validation failed.    
Additional properties not allowed: 2xx at #/paths///get/responses   JSON_OBJECT_VALIDATION_FAILED
```

Any idea?

maxidr: I checked the schema validator of Swagger (v2) and I see "^([0-9]{3})$|^(default)$".  This regex doesn't seem to support the "2xx".
I miss something?

climba03003: ```{0-9}XX``` is the syntax from OpenAPI V3 which is not allowed from Swagger Schema which use OpenAPI V2.

fox1t: @Eomm we need to make a clear distinction between v2 and v3. :(

I think it is safe to use 200 for this test.

maxidr: @Eomm this plugin supports OpenAPI v3?

mcollina: We use `2xx`  in Fastify for specifying those patterns. This module can receive `2xx` inside the schema defined for the routes.

climba03003: I found that a [bug](https://github.com/fastify/fastify-swagger/issues/250) report is exist for ```2XX``` response

Eomm: Thanks, @climba03003, I totally forgot that issue and I thought it was supported actually.. my bad

So I think the `2xx` should be fixed by another PR and this one is ok for the task it wants to accomplish 👍

Eomm (approved):

mcollina (approved):
lgtm

PR #305, opened 2020-10-08
fix: include swagger extensions if define at the root

### LINKED ISSUE #304: Swagger extension are not included if defined at the root
## 🐛 Bug Report

[Swagger extensions](https://swagger.io/docs/specification/2-0/swagger-extensions/) are not supported at the root of the spec

## To Reproduce

This code

```typescript
import fastifySwagger from "fastify-swagger";
fastify.register(fastifySwagger, { swagger: { "x-tension": "test" } });
```

throw a type error

```
Type '{ "x-tension": string; }' is not assignable to type 'Partial<Spec>'.
      Object literal may only specify known properties, and '"x-tension"' does not exist in type 'Partial<Spec>'
```

and I can verify that the extension is not included in the documentation.

## Your Environment

- *node version*: 12
- *fastify version*: ^3.4.1
- *os*: Mac

thomasthiebaud:
Fixes #304, but I'm not super fan of the implementation. 

I'm missing an update on the type definitions, but I can't express `Partial<SwaggerSchema.Spec>` and properties starting by `x-`. Do you have any preferences about what to do with that?

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -47,6 +47,13 @@ module.exports = function (fastify, opts, next) {
   const externalDocs = opts.swagger.externalDocs || null
   const transform = opts.transform || null
   const hiddenTag = opts.hiddenTag || 'X-HIDDEN'
+  const extensions = []
+
+  for (const [key, value] of Object.entries(opts.swagger)) {
+    if (key.startsWith('x-')) {
+      extensions.push([key, value])
+    }
+  }
 
   if (opts.exposeRoute === true) {
     const prefix = opts.routePrefix || '/documentation'
@@ -105,6 +112,10 @@ module.exports = function (fastify, opts, next) {
     }
     if (externalDocs) {
       swaggerObject.externalDocs = externalDocs
+    }
+
+    for (const [key, value] of extensions) {
+      swaggerObject[key] = value
     }
 
     const externalSchemas = Array.from(sharedSchemasMap.values())
--- a/test/swagger.js
+++ b/test/swagger.js
@@ -589,9 +589,9 @@ test('parses form parameters when all api consumes application/x-www-form-urlenc
 })
 
 test('includes swagger extensions', t => {
-  t.plan(3)
+  t.plan(5)
   const fastify = Fastify()
-  fastify.register(fastifySwagger, swaggerInfo)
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

Eomm (approved):
> Do you have any preferences about what to do with that?

I can't answer per TS, waiting @zekth

zekth:
One solution would be to set the swagger type to:
```ts
  swagger?: Partial<SwaggerSchema.Spec | Record<string,unknown>>;
```
Currently it's not possible but there is hope it will be in TypeScript 4.1. This example is not working yet but you get the idea.

Example: 
in playground: https://www.typescriptlang.org/play?target=99&ts=4.1.0-dev.20201001
```typescript
type prefixProperty<T extends string> = `${T}${string}`;
const test1: prefixProperty<'x-'> = 'x-ternal';
const test2: prefixProperty<'x-'> = 'ternal';
type prefixedObject<K extends prefixProperty<'x-'>,T> = { [P in K]:T}
const f: prefixedObject<prefixProperty<'x-'>,unknown> = {
    'x-bla': 'bar'
}
const d: prefixedObject<prefixProperty<'x-'>,unknown> = {
    'bla': 'bar'
}
const c: Record<prefixProperty<'x-'>,string> = {
    'x-bla': 'bar',
    'bla': 'bar'
}
```
Reference: https://github.com/microsoft/TypeScript/pull/40336

Mentionning @Ethan-Arrowood @fox1t @L2jLiga because we were talking about those templates earlier in fastify

fox1t:
@zekth, the playground link is not working.

Said that, as of today, since TS is a type system that doesn't check the vars' values, I would leave the types as is. I am not even sure that it is correct to introduce this type of check at compile time in the future.

L2jLiga:
I'm agree with @fox1t and also would keep as is

zekth:
fair point @fox1t

mcollina:
Any updates on this?

thomasthiebaud:
Do you want me to rebase? Based on the discussion I don't need to add anything to the current PR then

mcollina:
Could you just rebase? Thanks!

fox1t (approved):
LGTM

zekth (approved):
lgtm

PR #310, opened 2020-10-21
feature: when `stripBasePath` is set to false, do not strip basePath

andyburke:
Related to #232 #233

#### Checklist

### DIFF
--- a/dynamic.js
+++ b/dynamic.js
@@ -30,9 +30,13 @@ module.exports = function (fastify, opts, next) {
     })
   })
 
-  opts = opts || {}
-
-  opts.swagger = opts.swagger || {}
+  opts = Object.assign({}, {
+    exposeRoute: false,
+    hiddenTag: 'X-HIDDEN',
+    stripBasePath: true,
+    swagger: {},
+    transform: null
+  }, opts || {})
 
   const info = opts.swagger.info || null
   const host = opts.swagger.host || null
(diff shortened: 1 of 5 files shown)


dynamic.js
```diff
@@ -134,7 +134,7 @@ module.exports = function (fastify, opts, next) {
       const schema = transform
         ? transform(route.schema)
         : route.schema
-      let path = route.url.startsWith(basePath)
+      let path = route.url.startsWith(basePath) && !process.env.FASTIFY_SWAGGER_LEAVE_BASE_PATH
```
mcollina, line 137: Could you move this behind a plugin a plugin option?

mcollina (changes requested):

dynamic.js
andyburke: I'm not sure I understand. Are you asking to make this into a plugin or to make this controlled by a passed-in option?

andyburke: I've made some changes that I hope address what you were requesting. Please let me know what you think.

index.d.ts
```diff
@@ -46,6 +46,11 @@ export interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
   mode?: 'dynamic';
   swagger?: Partial<SwaggerSchema.Spec>;
   hiddenTag?: string;
+  /**
+   * Strips matching base path from routes in documentation
+   * @default true
+   */
+  stripBasePath?: boolean;
```
mcollina, line 53: We also have a test for the types, can you add this option there as well? Thanks

mcollina (commented):
good work! Just a minor thing

index.d.ts
andyburke: I pushed a small addition to the tests, but I'm not sure if it's what you were asking for.

mcollina (approved):
lgtm

Issue #316, opened 2020-10-30
Hide options are checked before transform


FriesFlorian:
## 🐛 Bug Report

The options to hide a route from the documentation are checked before the transform function is executed. 
This prevents being able to hide a route by adding `hide: true` or the hiddenTag to `tags` in transform()

## To Reproduce

```js
exposeRoute: true,
hiddenTag: "X-HIDDEN",
transform: schema => {
  return { ...schema, hide: true, tags: ["X-HIDDEN"] };
}
```

## Expected behavior

The above example should hide all routes.

From [`dynamic.js`](https://github.com/fastify/fastify-swagger/blob/master/dynamic.js):

```js
      if (route.schema && route.schema.hide) {
        continue
      }

      if (route.schema && route.schema.tags && route.schema.tags.includes(hiddenTag)) {
        continue
      }

      const schema = transform
        ? transform(route.schema)
        : route.schema
````

I think the two conditions should be executed after the schema transform() call

## Your Environment

- *node version*: 14.4.0
- *fastify version*: ^3.7.0
- *fastify-swagger*: ^3.1.2

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

FriesFlorian:
@mcollina just did that!

FriesFlorian:
Closing the issue since https://github.com/fastify/fastify-swagger/pull/317 has been merged

PR #317, opened 2020-10-30
fix: execute transform(schema) before checking schema.hide and hiddenTag

FriesFlorian:
Linked to issue https://github.com/fastify/fastify-swagger/issues/316

#### Checklist

- [x] run `npm run test`
- [x] commit message and code follows the guidelines

No need to update the doc as this is the expected behaviour of what's currently described in the section about hiding routes


mcollina (approved):
lgtm

zekth (approved):
lgtm. Good catch!

PR #328, opened 2020-12-08
Fix type of StaticDocumentSpec.document in index.d.ts

### LINKED ISSUE #327: Wrong Type for StaticDocumentSpec.document in index.d.ts
Code checks for an object
https://github.com/fastify/fastify-swagger/blob/6799cd424dfb02211e514c7b5257e5c57cea04ab/static.js#L54

but it is defined as a string. 
https://github.com/fastify/fastify-swagger/blob/6799cd424dfb02211e514c7b5257e5c57cea04ab/index.d.ts#L71

radzom:
Closes #327

The type of the property 'document' in the interface StaticDocumentSpec in the file 'index.d.ts' was formerly specified as 'string' and was corrected to object.

I think one could even specify it as one of
 - SwaggerSchema.Spec (for swagger V2 specs -already imported) or
 - OpenApi3Spec (for OpenApi V3 specs - where can this come from?)

Are there other versions that would then need to be supported?

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,6 @@
 import { FastifyPlugin } from 'fastify';
 import * as SwaggerSchema from 'swagger-schema-official';
+import { OpenAPIV2, OpenAPIV3 } from 'openapi-types';
 
 declare module 'fastify' {
   interface FastifyInstance {
@@ -68,7 +69,7 @@ export interface StaticPathSpec {
 }
 
 export interface StaticDocumentSpec {
-  document: string;
+  document: OpenAPIV2.Document | OpenAPIV3.Document;
 }
 
 export interface FastifyStaticSwaggerOptions extends FastifySwaggerOptions {
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "fs-extra": "^9.0.0",
     "joi": "^14.3.1",
     "joi-to-json-schema": "^5.1.0",
+    "openapi-types": "^7.0.1",
     "pre-commit": "^1.2.2",
     "standard": "^16.0.1",
     "swagger-parser": "^10.0.2",
--- a/test/types/http2-types.test.ts
+++ b/test/types/http2-types.test.ts
@@ -1,5 +1,6 @@
 import fastify from 'fastify';
 import fastifySwagger from '../..';
+import { minimalOpenApiV3Document } from './minimal-openapiV3-document';
 
 const app = fastify({
   http2: true
@@ -11,7 +12,7 @@ app.register(fastifySwagger, { transform: (schema : any) => schema });
 app.register(fastifySwagger, {
   mode: 'static',
   specification: {
-    document: 'path'
+    document: minimalOpenApiV3Document
   },
   routePrefix: '/documentation',
   exposeRoute: true,
--- a/test/types/imports.test.ts
+++ b/test/types/imports.test.ts
@@ -2,12 +2,13 @@ import fastify from "fastify";
 
 import swaggerDefault, { fastifySwagger, SwaggerOptions } from "../..";
 import * as fastifySwaggerStar from "../..";
+import { minimalOpenApiV3Document } from './minimal-openapiV3-document';
 
 const app = fastify();
 const fastifySwaggerOptions: SwaggerOptions = {
   mode: "static",
   specification: {
-    document: "path",
+    document: minimalOpenApiV3Document,
   },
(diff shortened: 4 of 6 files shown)


mcollina (approved):
lgtm

index.d.ts
```diff
@@ -68,12 +68,12 @@ export interface StaticPathSpec {
 }
 
 export interface StaticDocumentSpec {
-  document: string;
+  document: SwaggerSchema.Spec | object;
```
fox1t, line 71: Why also including `object` here?

radzom: Because otherwise OpenAPI3 specs would not be allowed. I am happy to integrate

radzom: So object is a catch all safety feedback in lack of knowing what specs have to be supported.
Just SwaggerSchema.Spec is not enough since it will only allow V2 Specs.
Hope i could make my point clear.

Ethan-Arrowood: Would rather us `Record<string, unknown>` instead of `object`. 

But also I don't see why we shouldn't use their official types package

radzom: So, which option is prefered?
A) change object to Record<string, unknown>
or
B) import types from "openapi-types"

I was not aware of that these types were official. Knowing that i strongly tend to option B.

examples/example-static-specification.json
```diff
@@ -8,10 +8,12 @@
       "email": "super.developer@gmail.com"
     }
   },
-  "servers": {
-    "url": "http://localhost:3000/",
-    "description": "Localhost (uses test data)"
-  },
+  "servers": [
```
radzom, line 11: This was an error found because of the new spec.

package.json
```diff
@@ -48,7 +48,8 @@
     "fastify-plugin": "^3.0.0",
     "fastify-static": "^3.3.0",
     "js-yaml": "^3.14.0",
-    "json-schema-resolver": "^1.2.0"
+    "json-schema-resolver": "^1.2.0",
+    "openapi-types": "^7.0.1"
```
radzom, line 52: added new package with ^ as all the other packages. (seems to be the prefered way here)

tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
radzom, line 1: I had to turn on the --resolveJsonModule switch for importing json files.
I first looked for commandline options in "tsd --help" but there are none.

package.json
mcollina: I would not recommend adding this dependency, maybe as a devDep?

tsconfig.json
```diff
@@ -0,0 +1,9 @@
...
+  "compilerOptions": {
+    "target": "es2015",
+    "module": "commonjs",
+    "strict": true,
+    "moduleResolution": "node",
+    "resolveJsonModule": true
```
mcollina, line 7: This requires adding some new settings to TS for using this lib, and I think it should be avoided.

mcollina (changes requested):

tsconfig.json
radzom: I can avoid this, by importing a minimal openapiv3 spec from a .ts file instead of a .json file.

package.json
radzom: good finding, already corrected

index.d.ts
fox1t: I prefer `openapi-types` path too. This plung supports only version 2, though. Are we introducing incompatibilities/breaking changes? If so I would go with `Record<string, unknown>`  for the moment.

fox1t (approved):

Ethan-Arrowood (approved):

mcollina (approved):
lgtm

radzom:
Hi, is there anything left I need, can or have to do for this pull request?
I am not impatient, just curious.
It is my first PR, so i am not sure if the Checklist needs to be checked and wo does it. 
My assumption was someone else :-).

Ethan-Arrowood:
@radzom nothing else on your end I believe. We do our merges and releases manually so whenever that happens for this plugin your work will be completed and released. Thank you for contributing 😄

climba03003:
@mcollina Sorry for tagging you. I do not know who is responsible for this plugin but I can see that there are no more change needed in here and with 3 persons approval. Should this PR be landed? I need the typing for #333.

mcollina:
Landed :)

PR #333, opened 2021-01-11
feat : add openapi 3 support

### LINKED ISSUE #265: Support for OpenAPI >= 3.0.0
## 🚀 Feature Proposal
I was wondering if there is any reason why we are using only v2 of Swagger spec. Therefore, I propose to add an option during the plugin registration to allow the users to choose between v2 and v3 when dynamic mode is used.

## Motivation
They are some useful features of OpenAPI v3 that are used by https://github.com/sinclairzx81/typebox to generate the awesome typed schemas that are used by Fastify routes.

### LINKED ISSUE #330: request params and querystring schema don't support allOf etc.
I have an schema of type 
```JS
{
  allOf: [
    {
      type: 'object',
      properties: [Object],
      required: undefined
    },
    {
      type: 'object',
      properties: [Object],
      required: undefined
    }
  ]
}
```
When i put it into either querystring or params schema for my endpoint the documentation generation crashes and i cannot access the docs page. Only shows error 'Failed to load API definition.'

However when i put the same schema into either body or headers schema, the documentations work properly.

climba03003:
Feedback are welcome while construction.

This PR used to add minimal support for OpenAPI 3. I do not assume complicated situation.

Issue related to OpenAPI 3
Resolve: #87, Resolve: #236, Resolve #265, Resolve #330
PR related to OpenAPI 3
Closes: #237

#### TODO List
- [x] refactor: reduce the number of files in root folder
- [x] refactor: extract common function in ```dynamic.js```
- [x] feat: add openapi 3 support
- [ ] feat: add translation layer between swagger 2 and openapi 3 (maybe in next PR, too much work required)

#### Checklist


mcollina:
huge +1 for the refactor.

darkgl0w:
Really nice to see a refactor and an upgrade in features for this package :D
I will follow your work on this PR and will be happy to review it ^^

climba03003:
Basic Functionality for OpenAPI 3 is added.
In order to make thing simple and not dealing with the transition between ```swagger 2.0``` and ```openapi 3.0```. I decide to add a new option ```openapi```. It may not be the ideal solution but it is the fastest way to get thing ready with semver-minor.

Maybe in the future, someone is willing to add a transition layer between ```swagger 2.0``` and ```openapi 3.0```.

climba03003:
TypeScript type will not updated for now. As #328 dealing with it.

mcollina:
@Eomm @SkeLLLa could you take a look?

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+    return propertiesMap
+  }
+
+  // $ref is in the format: #/definitions/<resolved definition>/<optional fragment>
+  const localReference = jsonSchema.$ref.split('/')[2]
+  return localRefResolve(externalSchemas[localReference], externalSchemas)
+}
+
+function stripBasePathByServers (path, servers) {
+  servers = Array.isArray(servers) ? servers : []
```
kibertoad, line 152: Wouldn't this lose the servers value if it's defined as a single entry?

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+  delete jsonSchema.type
+  delete jsonSchema.properties
+  delete jsonSchema.items
+  return jsonSchema
+}
+
+function localRefResolve (jsonSchema, externalSchemas) {
+  if (jsonSchema.type && jsonSchema.properties) {
+    // for the shorthand querystring/params/headers declaration
+    const propertiesMap = Object.keys(jsonSchema.properties).reduce((acc, h) => {
```
kibertoad, line 137: Full name for `headers` would be more easily readable

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+      const required = (jsonSchema.required && jsonSchema.required.indexOf(h) >= 0) || false
+      const newProps = Object.assign({}, jsonSchema.properties[h], { required })
```
kibertoad, line 139: Why use `Object.assign` instead of a simple spread?

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+function swagger2ParametersToOpenapi3 (jsonSchema) {
+  jsonSchema.schema = {}
+  jsonSchema.schema.type = jsonSchema.type
+  if (jsonSchema.type === 'object') {
+    jsonSchema.schema.properties = jsonSchema.properties
+  }
+  if (jsonSchema.type === 'array') {
+    jsonSchema.schema.items = jsonSchema.items
+  }
+  delete jsonSchema.type
```
kibertoad, line 128: `delete` is an expensive operation, any reason why setting to undefined wouldn't suffice here?

lib/util.js
```diff
@@ -0,0 +1,170 @@
+'use strict'
+
```
kibertoad, line 2: util.js sounds like a very generic name which is prone to become a God Object. maybe it could be split into more focused smaller files?

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+      }
+      done()
+    })
+  })
+
+  return {
+    routes,
+    Ref () {
+      const externalSchemas = Array.from(sharedSchemasMap.values())
+      return Ref({ clone: true, applicationUri: 'todo.com', externalSchemas })
```
kibertoad, line 32: Why the hardcoded `applicationUri`?

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+  const end = url.indexOf('/', ++start)
+
+  if (end === -1) {
+    return url.slice(0, start) + '{' + url.slice(++start) + '}'
+  } else {
+    return formatParamUrl(url.slice(0, start) + '{' + url.slice(++start, end) + '}' + url.slice(end))
+  }
+}
+
+function consumesFormOnly (schema) {
```
kibertoad, line 54: this is swagger-specific, likely would be helpful to separate OpenAPI stuff from Swagger stuff and from common stuff.

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+// For supported keys read:
+// https://swagger.io/docs/specification/2-0/describing-parameters/
+function plainJsonObjectToSwagger2 (container, jsonSchema, externalSchemas) {
+  const obj = localRefResolve(jsonSchema, externalSchemas)
+  let toSwaggerProp
+  switch (container) {
+    case 'query':
+      toSwaggerProp = function (properyName, jsonSchemaElement) {
+        jsonSchemaElement.in = container
+        jsonSchemaElement.name = properyName
```
kibertoad, line 73: properyName -> propertyName

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+          name: properyName,
+          required: jsonSchemaElement.required,
+          description: jsonSchemaElement.description,
+          type: jsonSchemaElement.type
+        }
+      }
+      break
+  }
+
+  return Object.keys(obj).reduce((acc, propKey) => {
```
kibertoad, line 113: I might be missing something, but isn't reduce an overkill here, isn't this a simple map operation?

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
...
+      if (cache.swaggerObject) return cache.swaggerObject
+    }
+
+    const swaggerObject = {}
+    let pkg
+
+    try {
+      pkg = JSON.parse(fs.readFileSync(path.join(__dirname, '..', 'package.json')))
+    } catch (err) {
+      return next(err)
```
kibertoad, line 41: There is a push to use `done` instead of `next` in fastify (see https://github.com/fastify/fastify/pull/2786), so it might make sense to rename here for consistency as well.

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    // Base Openapi info
+    // this info is displayed in the swagger file
+    // in the same order as here
+    swaggerObject.openapi = '3.0.0'
```
kibertoad, line 47: Why not 3.0.3?

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+const yaml = require('js-yaml')
+const { formatParamUrl, plainJsonObjectToSwagger2, swagger2ParametersToOpenapi3, stripBasePathByServers } = require('./util')
+
+module.exports = function (opts, routes, Ref, cache, next) {
```
kibertoad, line 8: this is a huge method that does many things. Would you be open to break it into several smaller functions, potentially spanning multiple files, to make it easier to follow the flow?

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
...
+    if (opts && opts.yaml) {
+      const swaggerString = yaml.safeDump(swaggerObject, { skipInvalid: true })
+      cache.swaggerString = swaggerString
+      return swaggerString
+    }
+
+    cache.swaggerObject = swaggerObject
+    return swaggerObject
+
+    function getBodyParams (parameters, body, consumes) {
```
kibertoad, line 198: Do we need to have all those functions defined in-line? Generally that causes issues for an optimizer.

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
...
+      add.forEach(_ => parameters.push(swagger2ParametersToOpenapi3(_)))
+    }
+
+    function getPathParams (parameters, path) {
+      const resolved = ref.resolve(path)
+      const add = plainJsonObjectToSwagger2('path', resolved, swaggerObject.components.schemas)
+      add.forEach(_ => parameters.push(swagger2ParametersToOpenapi3(_)))
+    }
+
+    function getHeaderParams (parameters, headers) {
```
kibertoad, line 224: These three functions look like one function with one additional parameter, no?

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
...
+      const resolved = ref.resolve(headers)
+      const add = plainJsonObjectToSwagger2('header', resolved, swaggerObject.components.schemas)
+      add.forEach(_ => parameters.push(swagger2ParametersToOpenapi3(_)))
+    }
+
+    // https://swagger.io/docs/specification/2-0/describing-responses/
+    function genResponse (fastifyResponseJson) {
```
kibertoad, line 231: full name `generateResponse` might read easier

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
...
+      // if the user does not provided an out schema
+      if (!fastifyResponseJson) {
+        return { 200: { description: 'Default Response' } }
+      }
+
+      const responsesContainer = {}
+
+      Object.keys(fastifyResponseJson).forEach(key => {
+        // 2xx is not supported by swagger
```
kibertoad, line 240: Can you elaborate?

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  return function (opts) {
+    if (opts && opts.yaml) {
+      if (cache.swaggerString) return cache.swaggerString
+    } else {
+      if (cache.swaggerObject) return cache.swaggerObject
+    }
+
+    const swaggerObject = {}
```
kibertoad, line 35: Why there are so many references to Swagger in OpenAPI implementation naming-wise? Is this intentional or just some copy-paste leftovers?

lib/dynamic.js
```diff
@@ -0,0 +1,36 @@
...
+    fastify.register(require('./routes'), { prefix })
+  }
+
+  const cache = {
+    swaggerObject: null,
+    swaggerString: null
+  }
+
+  if (Object.keys(opts.openapi).length > 0 && opts.openapi.constructor === Object) {
+    fastify.decorate('swagger', buildOpenapi(opts, routes, Ref, cache, next))
```
kibertoad, line 30: This doesn't look correct. Swagger is very much specifically OpenAPI 2.0, so using this name here seems misleading.

lib/openapi.js
```diff
@@ -0,0 +1,262 @@
...
+
+      Object.keys(fastifyResponseJson).forEach(key => {
+        // 2xx is not supported by swagger
+
+        const rawJsonSchema = fastifyResponseJson[key]
+        const resolved = ref.resolve(rawJsonSchema)
+
+        const content = {
+          'application/json': {}
+        }
```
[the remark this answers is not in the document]
SkeLLLa: Or `'*/*'` will make a trick.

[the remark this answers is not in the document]
kibertoad: `produces` is a Swagger thing, OpenAPI defines content-type as a part of `content` definition: https://swagger.io/docs/specification/describing-responses/

lib/dynamic.js
SkeLLLa: Maybe it's done for backward-compatibility, so there will be no need in major version bump.

lib/openapi.js
SkeLLLa: I think this one is swagger's legacy. Openapi does support `2XX`.

lib/dynamic.js
kibertoad: If that's the case, I would leave single `fastify.decorate` line but resolve actual value before calling it.
Don't think it would be breaking either way, as OpenAPI wasn't supported at all in the past.

lib/openapi.js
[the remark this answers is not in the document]
SkeLLLa: Yes, as well as consumes. But it's about fastify's schema and a way to define type of response.

lib/util.js
```diff
@@ -0,0 +1,170 @@
...
+  fastify.addHook('onRegister', async (instance) => {
+    // we need to wait the ready event to get all the .getSchemas()
+    // otherwise it will be empty
+    instance.addHook('onReady', (done) => {
+      const allSchemas = instance.getSchemas()
+      for (const schemaId of Object.keys(allSchemas)) {
+        if (!sharedSchemasMap.has(schemaId)) {
+          sharedSchemasMap.set(schemaId, allSchemas[schemaId])
+        }
+      }
```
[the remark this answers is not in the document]
Eomm: This is true.
The case to replicate would be:

- roo instance: register swagger plugin
- register a new context adding a schema.$id foo
- register a new context adding a different schema.$id foo

lib/util.js
Eomm: This is done to resolve relative $refs when the user (wrongly) doesn't set it.
Can't remember why I set it like this.. need to give it a check!

lib/util.js
climba03003: When I check the spec from swagger, servers can only accept array value.

kibertoad: Should we be silently ignoring invalid value then?

lib/util.js
climba03003: Most of the functions in ```util.js``` is extracted from the ```dynamic.js```. I do not really know what is the actual behavior or what is the goal of most of the functions.
For the ```Object.assign```, maybe it want to preserve the ```Symbol``` property.

lib/util.js
climba03003: ```delete``` is a must in this case, because swagger will check if the schema have invalid property or not. Setting to undefined will fail the test.

lib/openapi.js
climba03003: It use the swaggerObject inside, will passing a large object to everywhere have performance issue? If no, I will extract the function to other place.

kibertoad: Since objects are passed by reference without copying in JavaScript, size is irrelevant, it should be very cheap.

climba03003:
For the comment I have fixed, I marked the conversation as resolved. And the others need more detail or consensus.

lib/dynamic.js
climba03003: I think it have no difference between assigning the function before or after. It should use either one of them but not both.

For performance, I assume all the work should be done before ```ready``` event is fired. So, the user can call the right function on the ```fastify.ready```.

kibertoad: Yeah, I'm not suggesting anything drastic here, just for me this looks somewhat neater:

```
  const decoratorValue = resolveDecoratorValue(/*params*/)
  fastify.decorate('swagger', decoratorValue())
```

Could be a ternary operator, but judging by the length of condition, might get out of hand.

lib/openapi.js
climba03003: It is just copy and paste, I will rename them.

lib/openapi.js
climba03003: Forget to remove the invalid comment.

lib/openapi.js
climba03003: I think it is hard for the other to follow the flow when it because smaller components. The flow here is very strict forward. Maybe extract all the inline function will make it better.

kibertoad: It's less about branches and more about the difficulty of grasping many elements at once. Human brain is not really wired for that.
See https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two
and https://www.informit.com/articles/article.aspx?p=1398607 for more details on this topic.

climba03003: This maybe a little bit out of scope. For your provided link, we need a lot of refactor in order to achieve it, but this PR is mainly for adding the openapi 3 support.
I do the refactor first because most of the functions are commonly use between ```openapi``` and ```swagger``` but it should not be the goal in this PR.

lib/util.js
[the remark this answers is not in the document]
climba03003: I left comment to address the issue, it should be fixed in another PR. I think we should not put to much difference work in a single PR. It is hard to keep a good track and if there are problem, it is hard to revert.

lib/openapi.js
climba03003: I will left a comment to address the issue.

lib/util.js
climba03003: I will just left a comment to address the issue. It is not really related to this PR.

lib/util.js
climba03003: We need a document on this behavior.

SkeLLLa: Maybe it's worth to have some default value. E.g. like it's done here https://github.com/SkeLLLa/fastify-oas/blob/master/lib/openapi/constructor.js#L163

climba03003: The empty array can skip the for loop of strip base path. If we specify the default value and we know it will remove the leading ```/```, then we need to add the leading ```/``` back. Also, it is the function for strip the base path only. I will remain the empty array as it can do the work faster.

Maybe we need a better URL handling here instead of using ```URL```. 
```If the servers property is not provided, or is an empty array, the default value would be a Server Object with a url value of /.```

**Comment by climba03003** [comment]
@kibertoad @SkeLLLa @Eomm Sorry for tagging, I think it is ready for next branch of review. Do you have time to take a look on it?

File: lib/openapiUtil.js [hunk 21]
```diff
@@ -0,0 +1,90 @@
...
+    parameters[consume] = {
+      schema: bodyResolved
+    }
+  })
+}
+
+function getCommonParams (container, parameters, schema, ref, sharedSchema) {
+  const resolved = ref.resolve(schema)
+  const add = plainJsonObjectToSwagger2(container, resolved, sharedSchema)
+  add.forEach(_ => parameters.push(swagger2ParametersToOpenapi3(_)))
```
SkeLLLa, line 38: I think it will be better to give a proper name instead of just `_`.

lib/openapi.js
```diff
@@ -0,0 +1,194 @@
...
+          getBodyParams(swaggerMethod.requestBody.content, schema.body, schema.consumes, ref)
+        }
+
+        if (schema.params) {
+          getCommonParams('path', parameters, schema.params, ref, openapiObject.components.schemas)
+        }
+
+        if (schema.headers) {
+          getCommonParams('header', parameters, schema.headers, ref, openapiObject.components.schemas)
+        }
```
SkeLLLa, line 155: Openapi has also option to describe cookies: https://swagger.io/docs/specification/describing-parameters/#cookie-parameters.
Maybe it worth to add them as well. 
```js
        if (schema.cookie) {
          getCommonParams('cookie', parameters, schema.cookie, ref, openapiObject.components.schemas)
        }
```

lib/swaggerUtil.js
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // $ref is in the format: #/definitions/<resolved definition>/<optional fragment>
+  const localReference = jsonSchema.$ref.split('/')[2]
+  return localRefResolve(externalSchemas[localReference], externalSchemas)
+}
+
+// For supported keys read:
+// https://swagger.io/docs/specification/2-0/describing-parameters/
+function plainJsonObjectToSwagger2 (container, jsonSchema, externalSchemas) {
```
SkeLLLa, line 31: If cookie support will be added, it will require changes in here as well. But note that swagger2 doesn't support them.

lib/swaggerUtil.js
```diff
@@ -0,0 +1,139 @@
...
+  param.name = 'body'
+  param.in = 'body'
+  param.schema = bodyResolved
+  parameters.push(param)
+}
+
+function getCommonParams (container, parameters, schema, ref, sharedSchemas) {
+  const resolved = ref.resolve(schema)
+  const add = plainJsonObjectToSwagger2(container, resolved, sharedSchemas)
+  add.forEach(_ => parameters.push(_))
```
SkeLLLa, line 106: We could have better name than `_`

lib/swaggerUtil.js
climba03003: I have split the two method.

lib/openapi.js
climba03003: I will treat the input the same as ```querystring``` and we need to document it. As fastify it self do not support ```cookies``` field schema.

climba03003:
Oh, I shouldnt use rebase, it create a lot of useless commit. You can see the change start from ```refactor: better argument name``` c0439b6

mcollina:
I would recommend a squash.

climba03003:
@kibertoad @Eomm Do you have any time do look at it again. 
@SkeLLLa I dont know why GitHub cannot re-request review for you.

SkeLLLa (approved):

kibertoad (approved):

mcollina:
semver-major given the extent of the change?

jsumners:
> semver-major given the extent of the change?

I think that's a good idea.

mcollina:
This is missing some documentation update.

climba03003:
I will complete it today and also the typing update.

README.md
```diff
@@ -136,14 +137,29 @@ fastify.ready(err => {
...
+      tags: [ Object ]
     }
   }
   ```
 
   *All the above parameters are optional.*
-  You can use all the properties of the [swagger specification](https://swagger.io/specification/), if you find anything missing, please open an issue or a pr!
+  You can use all the properties of the [swagger specification](https://swagger.io/specification/v2/) and [openapi specification](https://swagger.io/specification/), if you find anything missing, please open an issue or a pr!
+
+  *Please note that when you specify `openapi` option, it will take precedence and ignore the `swagger` option.
```
mcollina, line 160:
```suggestion
  fastify-swagger will generate Swagger v2 by default. If you pass the `opeanapi` option it will generate OpenAPI instead.
```

README.md
```diff
@@ -136,14 +137,29 @@ fastify.ready(err => {
...
+      info: {
+        title: String,
+        description: String,
+        version: String,
+      },
+      externalDocs: Object,
+      servers: [ Object ],
+      components: Object,
+      security: [ Object ],
+      tags: [ Object ]
```
mcollina, line 152: Can you please comment out this block?

mcollina (changes requested):
I think we should throw if both `swagger` and `openapi` options are both set at the same time.

climba03003:
> 
> 
> I think we should throw if both `swagger` and `openapi` options are both set at the same time.

I think someone will left the `swagger` option there as backup and see the stability of the latest rollout of `openapi` support. Maybe I can add a warning instead of an error to notify the user swagger option will be ignored?

mcollina (approved):
lgtm

Eomm (changes requested):
Blocking because I need a couple of days to check the PR and I would like to review it 👍

Eomm (commented):
I think this feature is ok as the first step to OAS3 support

There is some issues with this usage:

<details>
  <summary>Click to expand!</summary>
```js
'use strict'

const Fastify = require('fastify')
const fastifySwagger = require('../../index')

const fastify = Fastify({ logger: true })
  openapi: {
    info: {
      title: 'Test swagger',
      version: '0.1.0'
    },
    servers: [{
      url: 'http://localhost'
    }],
    components: {
      securitySchemes: {
        apiKey: {
          type: 'apiKey',
          name: 'apiKey',
          in: 'header'
        }
      }
    }
  },
  exposeRoute: true
})

fastify.addSchema({
  $id: 'Order',
  type: 'object',
  properties: {
    id: {
      type: 'integer'
      // format: 'int64'
    }
  }
})

fastify.post('/', {
  schema: {
    body: {
      $ref: 'Order#'
    },
    response: {
      200: {
        $ref: 'Order#'
      }
    }
  }
}, () => {})

fastify.listen(8080, err => {
  console.log(err)
})

```
</details>

That lead to:

image

Because one big difference between oas2 and oas3 is the `components` over the old `definitions` keyword. (I use [this schema](https://editor.swagger.io/) as reference)

So I would land this PR as a major preview release to collect other edges cases before a major bump.

What do you think? @fastify/plugins

climba03003:
> 
> 
> I think this feature is ok as the first step to OAS3 support
> 
> There is some issues with this usage:
> Click to expand!
> ` 'use strict' const Fastify = require('fastify') const fastifySwagger = require('../../index') const fastify = Fastify({ logger: true }) fastify.register(fastifySwagger, { openapi: { info: { title: 'Test swagger', description: 'testing the fastify swagger api', version: '0.1.0' }, servers: [{ url: 'http://localhost' }], components: { securitySchemes: { apiKey: { type: 'apiKey', name: 'apiKey', in: 'header' } } } }, exposeRoute: true }) fastify.addSchema({ $id: 'Order', type: 'object', properties: { id: { type: 'integer' // format: 'int64' } } }) fastify.post('/', { schema: { body: { $ref: 'Order#' }, response: { 200: { $ref: 'Order#' } } } }, () => {}) `
> 
> `fastify.listen(8080, err => { console.log(err) }) `
> 
> That lead to:
> 
> image
> 
> Because one big difference between oas2 and oas3 is the `components` over the old `definitions` keyword. (I use [this schema](https://editor.swagger.io/) as reference)
> 
> So I would land this PR as a major preview release to collect other edges cases before a major bump.
> 
> What do you think? @fastify/plugins

I think it is the bug from `addHook`, it should be exist even for `swagger`. The main reason is that we never collect the schema from `addSchema` on `root` instance.

```JS
  routes.push(routeOptions)
})

fastify.addHook('onRegister', async (instance) => {
  instance.addHook('onReady', (done) => {
    const allSchemas = instance.getSchemas()  // we only collected the encapsulate schema
    for (const schemaId of Object.keys(allSchemas)) {
      if (!sharedSchemasMap.has(schemaId)) {
        sharedSchemasMap.set(schemaId, allSchemas[schemaId])
      }
    }
    done()
  })
})
```

Eomm (approved):
> it should be exist even for swagger.

Nope, it is managed by the (ugly sigh) `def-0` of the `json-schema-resolver` and it could be fixed there:
https://github.com/Eomm/json-schema-resolver/blob/39191491f3ef2c4b5fac5e19ad067459ccccdded/ref-resolver.js#L24
but your hotfix works 👍

climba03003:
> 
> 
> > it should be exist even for swagger.
> 
> Nope, it is managed by the (ugly sigh) `def-0` of the `json-schema-resolver` and it could be fixed there:
> https://github.com/Eomm/json-schema-resolver/blob/39191491f3ef2c4b5fac5e19ad067459ccccdded/ref-resolver.js#L24
> but your hotfix works 👍

I do not think it can be solved as you say as the `#/definitions` is hard coded in here:
https://github.com/Eomm/json-schema-resolver/blob/39191491f3ef2c4b5fac5e19ad067459ccccdded/ref-resolver.js#L113

What I said it exist for `swagger` is that, your example `addSchema` is on the root instance which we do not collect the schema currently and `json-schema-resolver` cannot resolve schema from an empty object. You can check my newly added test in here:
https://github.com/fastify/fastify-swagger/blob/35a9be8a793858c2094685ad1b8d45e7f7fef844/test/openapi.js#L785-L789
I need to wrap the `addSchema` in `register` in order to get the test work.
It should be a bug needed to file, so I will not fix it now.

SkeLLLa:
For schemas there's slightly different approach that might be more elegant solution rather than resolve it. I've described it here: https://github.com/SkeLLLa/fastify-oas/issues/30#issuecomment-632641407 but it require lot's of work, so it should be different PR.

index.d.ts
```diff
@@ -1,4 +1,5 @@
-import { FastifyPlugin } from 'fastify';
+import { FastifyPluginCallback } from 'fastify';
+import { OpenAPI, OpenAPIV2, OpenAPIV3 } from 'openapi-types';
```
danez, line 2: Because of this openapi-types needs to be a dependency and not dev-dependency.
See #340

climba03003: Actually, it is landed from here #328

danez: Bit this pr was landed after #328 if i see that correctly here. So on master index.d.ts has a dependency on `openapi-types`, but it does not list it as dependency.

climba03003: This PR is merging the #328 by fast-forwarding commit. Actually, I do not touch `index.d.ts` by my own.
`4.0.0` indeed is merging two PR. #328 and #333, you can check the commit history in here to see the detail.

climba03003: Let say in this way, even without this PR, #328 will cause the same problem. If the problem is adding `openapi-types` as devDeps. It is added in #328.

PR #339, opened 2021-01-27
refactor: increase readability

climba03003:
This PR is refactor only which goal is to increase the readability of code as @kibertoad advise and better organize the file.
It should not add any new feature inside this PR. 

#### Checklist


index.js
```diff
@@ -2,19 +2,21 @@
...
       break
-    case 'dynamic':
-      setup.dynamic(fastify, opts, next)
+    }
+    case 'dynamic': {
+      const setup = require('./lib/mode/dynamic')
+      setup(fastify, opts, next)
       break
+    }
     default:
```
jsumners, line 20: I do prefer to make the `case` blocks explicit with `{...}`. Since you added them to the other cases, it would be good to add them to this one as well.

lib/util/openapi.js
```diff
@@ -0,0 +1,235 @@
+'use strict'
```
jsumners, line 1: If these utility functions are for use only within the openapi spec, why not put this file in `lib/spec/openapi/`?

lib/util/swagger.js
```diff
@@ -0,0 +1,238 @@
+'use strict'
```
jsumners, line 1: Same question as the openapi utility file.

index.d.ts
```diff
@@ -1,14 +1,13 @@
 import { FastifyPluginCallback } from 'fastify';
 import { OpenAPI, OpenAPIV2, OpenAPIV3 } from 'openapi-types';
-import * as SwaggerSchema from 'swagger-schema-official';
```
mcollina, line 3: #Could you please remove this from package.json and add openapi-types to dependencies?

mcollina (changes requested):

lib/util/openapi.js
climba03003: It is just my personal preference, no special reason for it. I will follow your guide to combine it in single file.

index.d.ts
climba03003: Done

index.js
climba03003: Done

lib/util/openapi.js
climba03003: Done

lib/util/swagger.js
climba03003: Done

mcollina (approved):
lgtm

lib/util/openapi.js
jsumners: I wasn't suggesting a single file. I was suggesting something like `lib/spec/openapi/{index.js,utils.js}`.

jsumners (approved):
LGTM

But FWIW, I was suggesting `lib/spec/openapi/{index.js,utils.js}`. Not adding all of the utility functions into the same file as the implementation. I'm fine either way though.

climba03003:
> 
> 
> LGTM
> 
> But FWIW, I was suggesting `lib/spec/openapi/{index.js,utils.js}`. Not adding all of the utility functions into the same file as the implementation. I'm fine either way though.

I think I missed the trailing slash in your comment, so I think you are suggesting combining into a single file. 
I do love to split the `utils` function in separate files and I am going for this approach.

jsumners (approved):
LGTM

Issue #336, opened 2021-01-26
Support providing response description different from response body description


jeremy-w:
## 🚀 Feature Proposal

When `responseDescription` is present in the response schema, then move it up to the response's `description` field. If not present, then proceed with the existing behavior of duplicating the `description` to the response `description`.

## Motivation

I want to separately document the response itself vs the response body. This is especially handy when reusing the payload schema.

## Example

```js
        200: {
          responseDescription: 'Successfully authenticated',
          description: 'Always the same response body. TBH I have no clue why it even sends this object in response. The status code is enough. Go figure.',
          type: 'object',
          properties: {
            authorization: {
              type: 'string',
              enum: ['success'],
            },
          },
        }
```

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

jeremy-w:
@mcollina I have this working for Swagger. I still need to do similarly for OpenAPI, then update the docs.

For docs, should I touch anything beyond the README?

mcollina:
> For docs, should I touch anything beyond the README?

No, that's enough

climba03003:
Resolved by #341

PR #341, opened 2021-01-27
Transform Fastify schema.responseDescription to OAS response.description  [#336]

jeremy-w:
## Before
Any description provided in a response schema was also used for the response itself.

For example, this route:

```js
fastify.get('/description', {
  schema: {
    response: {
      200: {
        description: 'response and schema description',
        type: 'string'
      }
    }
  }
}, () => {})
```

generated this schema:

```json
{
  "/description": {
    "get": {
      "responses": {
        "200": {
          "description": "response and schema description",
          "schema": {
            "description": "response and schema description",
            "type": "string"
          }
        }
      }
    }
  }
}
```

## Now
The existing behavior stands, but you can also provide an independent description of the response via the `responseDescription` field.

For example, this route:

```js
    fastify.get('/responseDescription', {
      schema: {
        response: {
          200: {
            responseDescription: 'description for the response as a whole',
            description: 'description for the response body',
            type: 'string'
          }
        }
      }
    }, () => {})
```

generates this schema:

```json
{
  "/responseDescription": {
    "get": {
      "responses": {
        "200": {
          "description": "response description",
          "schema": {
            "description": "schema description",
            "type": "string"
          }
        }
      }
    }
  }
}
```

Finishes #336.

#### Checklist

- [X] run `npm run test` and `npm run benchmark` (NOTE: this project has no `benchmark` package script)
- [X] updates applied to Swagger schema generation
- [x] updates applied to OpenAPI schema generation

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ fastify.ready(err => {
 
   fastify-swagger will generate Swagger v2 by default. If you pass the `opeanapi` option it will generate OpenAPI instead.
 
-  Example of the `fastify-swagger` usage in the `dynamic` mode, `swagger` option is available [here](examples/dynamic-swagger.js) and `openapi` option is avaiable [here](examples/dynamic-openapi.js).
+  Example of the `fastify-swagger` usage in the `dynamic` mode, `swagger` option is available [here](examples/dynamic-swagger.js) and `openapi` option is available [here](examples/dynamic-openapi.js).
 
 ##### options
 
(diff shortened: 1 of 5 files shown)


jsumners:
I'm confused. How does the spec standard `description` not provide this?

jeremy-w:
@jsumners the end result of this PR is to copy up responseDescription to the spec-standard ResponseObject.description while deleting it from the SchemaObject.

FastifySchema has a one-level response schema focused on serializing the response body. WRT Swagger, that spec is doubling as both the ResponseObject and the ResponseObject's SchemaObject. currently any provided description gets duped to both of those objects.

to experience this issue for yourself, try to write a `fastify.route` options schema that will generate a schema where `paths['/'].get.responses['200'].description` is not the same as `paths['/'].get.responses['200'].schema.description`.

jsumners:
> to experience this issue for yourself, try to write a `fastify.route` options schema that will generate a schema where `paths['/'].get.responses['200'].description` is not the same as `paths['/'].get.responses['200'].schema.description`.

I'm assuming your example in the top level description of this PR shows what you're suggesting. If not, please provide an example so that I do not have to guess at getting it right.

In your example, I do not understand the difference between "description for the response as a whole" and "description for the response body". How are they not the same thing?

jeremy-w:
**example:** I’ll console log example generated schemas from the tests next I work on this. The example in the PR description shows the code written, not generated.

**not the same thing:** They are distinct description fields.

The response object has a required description: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-10

And the schema object has its own description inherited from JSON schema: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema-object

As to why, I can only speculate: Just as I can describe an envelope (pink, has stickers) separately from its contents (card, has glitter), it can be useful to describe a response itself (takes 10 secs to generate, expect delays) separately from its body (a number).

jsumners:
@mcollina I'm not sure how useful my review will be here.

jeremy-w:
I totally missed you were set as reviewer. GitHub’s mobile UI doesn’t do much to call it out. It is definitely early to review; this is about half done. It should make a lot more sense once the docs are written. 😅

mcollina:
cc @climba03003 can you take a look?

jeremy-w:
I've now added documentation to the README in the "dynamic" section. I've included the core examples as "Before" and "Now" examples in the PR description as requested by @jsumners. I hope the examples clear up any confusion.

climba03003 (approved):
I know that sometimes people like to separate the `response description` and `schema description` which can provide a better documentation.
For example the response schema is shared across multiple end-point. `schema description` can be `Generic Response Schema` and the `response description` should descripted per end-point like `Return Foo` or `Return Bar`.

It LGTM for adding this feature.

test/spec/openapi/schema.js
```diff
@@ -40,3 +40,73 @@ test('suport - oneOf, anyOf, allOf', t => {
...
+            type: 'string'
+          }
+        }
+      }
+    }, () => {})
+    await fastify.ready()
+
+    // When the Swagger schema is generated
+    const swaggerObject = fastify.swagger()
+    console.log('responseDesc', require('util').inspect(swaggerObject, false, 999))
```
climba03003, line 99: Can you remove the `console.log` here, I think it is used for debugging the test only.
```suggestion
```

climba03003 (changes requested):

test/spec/openapi/schema.js
jeremy-w: Whoops. That was there when I was copying content from test schemas into README. Thanks for catching it!

jeremy-w:
Rebased to resolve conflicts. 👍

climba03003 (approved):
LGTM

README.md
```diff
@@ -173,6 +172,183 @@ fastify.ready(err => {
...
+```
+
+If you want to provide a different description for the response as a whole, instead use the `responseDescription` field alongside `description`:
+
+```js
+fastify.get('/responseDescription', {
+  schema: {
+    response: {
+      200: {
+        responseDescription: 'response description',
```
Eomm, line 299: I'm not pro to introduce a new keyword in the JSON-Schema spec.

What we are trying to solve is:

- when the user needs to set as input or output a "non-object" value it is impossible to define two descriptions: one for the model and one for the field itself.

So, to archive it I would use `Symbols` in order to avoid the `responseDescription` keyword, doing so:

- it is not necessary to do the shallow copy of the object with the `delete`
- the standard is preserved

The cons is that the user can't put the schema in a simple `json` - but it is required a `.js` file

Other thought?

Eomm (changes requested):

README.md
jeremy-w: The generated schema has no new keywords, but we lose the ability to describe a response body with a responseDescription field (unless we were to check it’s a string before messing with it), so I think you’re onto something.

Would if be better to use a symbol or an “x-…” property, do you think? Or just make sure responseDescription is a string before transforming it?

I’m not sure the shallow copy is required. But making it avoids possibly getting that analysis wrong.

jsumners: > I'm not pro to introduce a new keyword in the JSON-Schema spec.

This was my concern as well. However, the spec doesn't _prohibit_ arbitrary keywords. It just leaves interpretation of those keywords up to the interpreter implementation. So I was trying to understand the use case before making this argument.

I am not clear that any other spec, built on top of JSON Schema (Swagger/OpenAPI), provides this `responseDescription` property, though. So I think this PR is introducing completely new syntax.

jeremy-w: i repeat: the JSON schema has no new keywords. just as a YAML schema might merge in properties and use references that are transformed away in generating the final JSON schema, similarly the schema supplied to Fastify, when processed into the JSON schema by fastify-swagger, has no `responseDescription` field; this is an artefact of the input format only, not of the actual schema.

ultimately, i am waiting on direction here:

- use a symbol?
- use an x- prefix?
- proceed as-is?

which one will be acceptable for merge? i'm not beholden to any one of the options: i just want some way to provide a value for the response.description field in the generated Swagger/OAS spec.

jsumners: @jeremy-w This comment thread is attached to an example showing _exactly_ that this requires a new keyword added to the JSON Schema provided to the Fastify route configuration. It is entirely possible that this could conflict with JSON Schema interpreters. Therefore, we must do due diligence to understand the desire versus the implications.

climba03003: I think we should take a assumption, if people need to use this `plugin` properly or take the full functionality of this `plugin`.
1. They should specify the whole `json schema` which means `type` keyword must be exist. Then, we can check it in `onRoute` hooks, before the `schema` being mutated.
2. We may also `warn` the user if they do not specify `type` keyword, some functionality will not work properly.

Issue related to ppl missing the `type` keyword and causing the abnormal behavior.
#347

Eomm: > use an x- prefix?

👍 I think this would be a good compromise to archive the feature and avoid conflicts with the spec and other tools as well

jeremy-w: thank @Eomm. i've updated this so the key is now `x-response-description`.

this is in line with the [OAS 2 guidance for specification extensions](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#specification-extensions). Since it doesn't start with `x-oas-`, it is also in line with that of [OAS 3](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#specificationExtensions).

the `x-response-description` key (as well as the previous `responseDescription` key) appear to be in line with the [JSON schema guidance on extensions](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.6.5), which can be roughly summarized as "ignore keywords you don't recognize, but don't expect others to magically support your weird custom keywords".

mcollina (approved):
lgtm

jeremy-w:
@Eomm does this look good to you?

Eomm (approved):

PR #345, opened 2021-01-30
Fix swagger nxx response

climba03003:
Resolve: #250

Caption from [swagger](https://swagger.io/docs/specification/describing-responses/)
`
To define a range of response codes, you may use the following range definitions: 1XX, 2XX, 3XX, 4XX, and 5XX. If a response range is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code.
`

I do not introduce a new option as @Eomm suggestion. It will also not throwing error because it may break someone code.

Behavior for Swagger 
- When I see `Nxx` status code, transform it to `200` status code.
- If we find conflict of `200` status code, drop the transformed one as it is more general response.
- Append the transformed status code to response.

Behavior for OpenAPI
- All `Nxx` status code will enforce upper case.
- Append the transformed status code to response.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,34 @@ fastify.ready(err => {
  | openapi       | {}       | Openapi configuration.                |
  | transform     | null     | Transform method for schema.          |
 
+##### 2XX status code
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

mcollina (changes requested):

mcollina (approved):
lgtm

Issue #335, opened 2021-01-26
Enable documenting response headers


jeremy-w:
## 🚀 Feature Proposal

Transform a Fastify schema response's `headers` field into the Swagger [response object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#responseObject)'s `headers` field.

## Motivation

I want to document a response header.

## Example

Given this route declaration:

```js
  fastify.route({
    method: 'GET',
    url: '/cookie',
    schema: {
      operationId: 'getCookie',
      summary: 'You get a cookie.',
      response: {
        '200': {
          description: 'Everyone loves cookies.',
          type: 'object',
          properties: {
            delicious: { type: 'boolean', default: 'true' },
          },
          headers: {
            'Set-Cookie': {
              type: 'string',
            },
          },
        },
      },
    },
    handler: () => {
      return
    },
  })
```

It would generate a [paths object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#paths-object) like:

```json
"/cookie": {
    "get": {
        "operationId": "getCookie",
        "summary": "You get a cookie.",
        "responses": {
            "200": {
                "headers": {
                    "Set-Cookie": {
                        "type": "string"
                    }
                },
                "schema": {
                    "description": "Everyone loves cookies.",
                    "type": "object",
                    "properties": {
                        "delicious": {
                            "type": "boolean",
                            "default": "true"
                        }
                    }
                },
                "description": "Everyone loves cookies."
            }
        }
    }
},
```

But currently it generates:

```json
"/cookie": {
    "get": {
        "operationId": "getCookie",
        "summary": "You get a cookie.",
        "responses": {
            "200": {
                "schema": {
                    "description": "Everyone loves cookies.",
                    "type": "object",
                    "properties": {
                        "delicious": {
                            "type": "boolean",
                            "default": "true"
                        }
                    },
                    "headers": {
                        "Set-Cookie": {
                            "type": "string"
                        }
                    }
                },
                "description": "Everyone loves cookies."
            }
        }
    }
},
```

which is not considered valid Swagger 2.

## Notes

I think the needed change would be to [the assignment in `genResponse` in dynamic.js](https://github.com/fastify/fastify-swagger/blob/f1c441b4662225499f084bbd33dd6aa096212ac8/dynamic.js#L303-L307).

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

[...]

Issue #313, opened 2020-10-28
Response without a body is unsupported


pelepelin:
## 🐛 Bug Report

Swagger 2 allows to specify a response without a body by omitting `schema` field. 

https://swagger.io/specification/v2/#responseObject
> If this field does not exist, it means no content is returned as part of the response.

fastify + fastify-swagger do not provide a way to specify response without a body. Note that a response with empty object is not the same as no body. This is most important when used with [204 status code](https://tools.ietf.org/html/rfc7231#section-6.3.5) which must not contain a body.

## To Reproduce

```js
const schema = {
  response: {
    204: {
      description: 'Success'
    }
  }
}
```

Gives fatal error
```
Could not start server: FastifyError [FST_ERR_SCH_SERIALIZATION_BUILD]: Failed building the serialization schema for POST: /url, due to error schema is invalid: data.properties['description'] should be object,boolean
```
And
```js
const schema = {
  response: {
    204: {
      type: "object",
      description: 'Success'
    }
  }
}
```

produces Swagger schema which assumes `{}` (empty object) body.
```json
"responses":{"204":{"schema":{"type":"object","description":"Success"},"description":"Success"}}
```

And
```js
const schema = {
  response: {
    204: {
      type: "null",
      description: 'Success'
    }
  }
}
```

produces Swagger schema which is still not equal to no content
```json
"responses":{"204":{"schema":{"type":"null","description":"Success"},"description":"Success"}}}
```

## Expected behavior

Given the schema

```js
const schema = {
  response: {
    204: {
      description: 'Success'
    }
  }
}
```

Should produce Swagger schema
```json
"responses":{"204":{"description":"Success"}}
```

## Your Environment

- *node version*: 12
- "fastify": "3.7.0",
- "fastify-swagger": "3.4.0",
- *os*: Linux

## Related

#13 fist report, different error message
#19 unfinished PR
#115 report, duplicate of #13
#116 unfinished PR
fastify/fastify#609 claims this is fixed, but apparently it's not, just the error message is different. I guess that without explicit `type` fastify serialization treats "description" as a field declaration, which it is not.

Both #19 and #116 try to fix the issue using explicitly specified `type: 'null'`, which IMHO is not ideal because it should not be required unless there is no other way to disable JSON validation and serialization for such responses as they're in fact not JSON.

pelepelin:
[...]

PR #346, opened 2021-02-01
feat: response headers and empty body

climba03003:
This PR used to fix the existing bug of handling empty body and add the function for setting response header.
Resolve: #313
Resolve: #335

It is a little bit bulky in `openapi` implementation, as we need to transform most of the schema.

Header Behavior for Swagger:
- add `headers` in response
- remove `headers` from schema 

Header Behavior for OpenAPI:
- add `headers` in response which need to do some `transformation`
- remove `headers` from schema

Note: we must use the keyword `delete` in here because `swagger.validate` itself will validate invalid field even you set it as `undefined`.

Empty Body Behavior:
- we only add `schema` into `response` when status code is not equal to 204

Note: we need to cast status code to `string` to ensure the check pass.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,24 @@ Example:
 }
 ```
 
+##### response headers
+You can decorate your own response headers by follow the below example.
+```js
+{
+  response: {
+    200: {
+      type: 'object',
+      headers: {
+        'X-Foo': {
+          type: 'string'
+        }
+      }
+    }
+  }
+}
+```
+Note: You need to specify `type` property when you decorate the response headers, otherwise the schema will be modify by `fastify`.
+
 ##### static
  `static` mode should be configured explicitly. In this mode `fastify-swagger` serves given specification, you should craft it yourself.
   ```js
--- a/README.md
+++ b/README.md
@@ -219,6 +219,19 @@ You can decorate your own response headers by follow the below example.
 ```
 Note: You need to specify `type` property when you decorate the response headers, otherwise the schema will be modify by `fastify`.
 
+##### status code 204
+We support status code 204 and return empty body. Please specify `type: 'null'` for the response otherwise `fastify` itself will fail to compile the schema.
+```js
+{
+  response: {
+    204: {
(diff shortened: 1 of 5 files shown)


climba03003:
Side effect (Breaking):
It is more or less the similar case for `description` only response schema in `fastify`.

If anyone specify the response like the following,
```JS
{
   headers: { type: 'object' }
}
```
which is valid in `fastify` and `fastify` will treated as below.
```JS
{
  type: 'object',
  properties: {
    headers: { type: 'string' }
  }
}
```

However, we are hardly determine the field name `headers` is belong to `properties` or `keyword`. Should we just specify it in the document? or we do not implement this feature?

mcollina:
> We are hardly determine the field name headers is belong to properties or keyword. Should we just specify it in the document? or we do not implement this feature?

Both are valid approaches, whatever you prefer. If we detect this, we should log something as warn.

climba03003:
It seems like `fastify` itself will modify the route options.
I can only get the below schema after `fastify.ready`.
```
{
  type: 'object',
  properties: {
    headers: { type: 'object' }
  }
}
```
So, I do not need to care about the conflict problem in code. But I should document this behavior.

mcollina (approved):
lgtm

jeremy-w:
> fastify itself will modify the route options

would the symbol approach suggested for smuggling in responseDescription work better with how fastify handles this? (i'm hoping it might ignore it / not inject it into properties.)

gyszalai:
It seems the fix is not correct. In case of

```js
response: {
  204: {
    type: 'null',
    description: 'my description'
  }
}
```

yields this in the swagger file:
```json
"404":{
  "description": "my description",
  "schema": {
     "description": "my description",
     "type": "null"
  }
}
```

gyszalai:
the `schema` property should not be present in the swagger descriptor

mcollina:
I would recommend to open a new issue with a full repro, or a straight PR.

Issue #228, opened 2020-02-14
Supply additional configuration to swagger-ui


Boshen:
## 🚀 Feature Proposal

Swagger ui can be modified by supplying  additional configuration parameters.
It's documented [here](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md)

As far as I can tell, The `new SwaggerUI({...})` call is hard coded in static/index.html,
so there is currently no options to do this in fastify-swagger.

By looking at swagger-ui-express, they served a modified html, see [here](https://github.com/scottie1984/swagger-ui-express/blob/master/index.js#L120).
It this something we can do too?

mcollina:
A PR that implemented that would be awesome!

Il giorno ven 14 feb 2020 alle ore 17:34 Boshen <notifications@github.com>
ha scritto:

gustawdaniel:
Especially if `css` customization would be possible

https://github.com/fastify/fastify-swagger/issues/82

giovanni-bertoncelli:
Any update on this?

mcollina:
@giovanni-bertoncelli Would you like to send a Pull Request to address this issue? Remember to add unit tests.

giovanni-bertoncelli:
I can try :)

giovanni-bertoncelli:
@mcollina There you go

PR #351, opened 2021-02-02
feat: pass swagger-ui additional configuration

giovanni-bertoncelli:
## Porpouse

This PR allows to pass a dynamic configuration to the generated swagger UI. The option that should be used in the plugin configuration in order to accomplish that is `uiConfig`.

Configuration docs: https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md

## Implementation

Since there is an issue configuring `swagger-ui-dist` (that distributes only static files, see https://github.com/swagger-api/swagger-ui/issues/5710) without changing the content of the static assets, the only way to configure the UI with a dynamic external configuration is by editing the `index.html` file and the script within that loads the swagger ui in the page.

There were two approaches:
1. Intercepting the `index.html` request on the documentation routes and distributing an altered html file that load swagger UI with the configuration merged with the one passed by the user.
2. Editing the file `index.html` when the project gets prepared (`npm run prepare`) in order to retrieving a remote configuration fetched through an XHR request on a designated endpoint.

Since the index.html file was already edited in the prepare stage I went for the second approach. Now a new route under `GET documentation/uiConfig` gets declared and returns the final ui configuration and the script inside the index page fetches the configuration from that endpoint. Finally the page loads the swagger ui from the given configuration.

## Drawbacks

Since there is no way to serialize functions over HTTP and JSON there is no way to pass functions to the parameters of the `uiConfig`, so only string/object/number configuration keys will be effective. Neither the first alternative implementation allowed to do that (at least without writing the plain function in the HTML file). However I think this is a fair compromise.

## Related issues

- #228 
- [swagger-ui#5710](https://github.com/swagger-api/swagger-ui/issues/5710)

## Others

The `npm run test` didn't run all the test inside the folder `test/` since the regex in the command was working only at the first directory level. Now it runs all the test till the 3rd directory level.

#### Checklist

### DIFF
--- a/lib/mode/dynamic.js
+++ b/lib/mode/dynamic.js
@@ -16,7 +16,8 @@ module.exports = function (fastify, opts, done) {
 
   if (opts.exposeRoute === true) {
     const prefix = opts.routePrefix || '/documentation'
-    fastify.register(require('../routes'), { prefix })
+    const uiConfig = opts.uiConfig || {}
+    fastify.register(require('../routes'), { prefix, uiConfig })
   }
 
   const cache = {
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "scripts": {
     "prepare": "node lib/util/prepare-swagger-ui",
     "prepublishOnly": "npm run prepare",
-    "test": "standard && tap --100 test/*.js test/**/*.js test/**/**/*.js && npm run typescript",
-    "test:ci": "standard && tap test/*.js test/**/*.js test/**/**/*.js --coverage-report=lcovonly && npm run typescript",
+    "test": "standard && tap --100 'test/**/*.js' && npm run typescript",
+    "test:ci": "standard && tap 'test/**/*.js' --coverage-report=lcovonly && npm run typescript",
     "typescript": "tsd"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "scripts": {
     "prepare": "node lib/util/prepare-swagger-ui",
     "prepublishOnly": "npm run prepare",
-    "test": "standard && tap --100 test/**/*.js && npm run typescript",
-    "test:ci": "standard && tap test/**/*.js --coverage-report=lcovonly && npm run typescript",
+    "test": "standard && tap --100 test/*.js test/**/*.js test/**/**/*.js && npm run typescript",
+    "test:ci": "standard && tap test/*.js test/**/*.js test/**/**/*.js --coverage-report=lcovonly && npm run typescript",
     "typescript": "tsd"
   },
   "repository": {
@@ -41,7 +41,7 @@
     "standard": "^16.0.1",
     "swagger-parser": "^10.0.2",
     "swagger-ui-dist": "3.41.1",
-    "tap": "^14.10.8",
+    "tap": "^14.11.0",
     "tsd": "^0.14.0"
   },
   "dependencies": {
--- a/test/route.js
+++ b/test/route.js
@@ -115,7 +115,7 @@ test('/documentation/uiConfig route', t => {
     t.error(err)
 
     const payload = JSON.parse(res.payload)
-    
+
     t.match(payload, uiConfig, 'Invalid uiConfig received')
   })
 })
--- a/test/route.js
+++ b/test/route.js
@@ -116,7 +116,7 @@ test('/documentation/uiConfig route', t => {
 
     const payload = JSON.parse(res.payload)
 
-    t.match(payload, uiConfig, 'Invalid uiConfig received')
+    t.match(payload, uiConfig, 'uiConfig should be valid')
   })
 })
 
--- a/test/route.js
+++ b/test/route.js
@@ -86,6 +86,40 @@ test('/documentation/json route', t => {
   })
 })
 
+test('/documentation/uiConfig route', t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  const uiConfig = {
+    docExpansion: 'full'
+  }
+
+  const opts = {
+    ...swaggerOption,
+    uiConfig
+  }
+
+  fastify.register(fastifySwagger, opts)
+
+  fastify.get('/', () => {})
+  fastify.post('/', () => {})
+  fastify.get('/example', schemaQuerystring, () => {})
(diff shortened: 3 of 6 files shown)


giovanni-bertoncelli:
The final configuration passed to the Swagger UI is composed by merging sequentially:

- The default swagger UI configuration 
- The user defined configuration
- The `{url, oauth2RedirectUrl}` configuration

Fdawgs (commented):
Commit message for 9d96c34 should be `test`, not `feat`

test/route.js
```diff
@@ -86,6 +86,40 @@ test('/documentation/json route', t => {
...
+    method: 'GET',
+    url: '/documentation/uiConfig'
+  }, (err, res) => {
+    t.error(err)
+
+    const payload = JSON.parse(res.payload)
+
+    t.match(payload, uiConfig, 'Invalid uiConfig received')
+  })
+})
```
mcollina, line 121: Can you please add a test with a valid uiConfig as well?

mcollina (commented):
Good work!

test/route.js
giovanni-bertoncelli: The configuration passed in that test is actually valid. Maybe I have to change the test message?

mcollina: Sorry about it, I misunderstood the assertion.

mcollina (approved):
lgtm

package.json
```diff
@@ -6,8 +6,8 @@
   "scripts": {
     "prepare": "node lib/util/prepare-swagger-ui",
     "prepublishOnly": "npm run prepare",
-    "test": "standard && tap --100 test/**/*.js && npm run typescript",
-    "test:ci": "standard && tap test/**/*.js --coverage-report=lcovonly && npm run typescript",
+    "test": "standard && tap --100 test/*.js test/**/*.js test/**/**/*.js && npm run typescript",
```
climba03003, line 9: I think this change is not needed as `test/**/*.js` means any files under `test` folder.

climba03003 (changes requested):

package.json
giovanni-bertoncelli: I used to think that too but instead running with the npm it does not consider deeper level files...
As a notice: in fact the coverage shown in the repository badge is currently 46% instead of 100%...

climba03003: I face the same issue when I do a clean install of packages. It seems like `tap` changes the behavior of glob.

climba03003 (approved):
LGTM

I get stuck in your option 1 when I see this feature request. My other consideration was CSP, but updating the `index.html` in `npm run prepare` is a good solution without affecting CSP.
Good Approach.

climba03003:
Just a suggest, will you consider replacing the whole block of `<script></script>` in `index.html`? I think you already changed all the content inside it.
It is more simple to use a single replace.

package.json
HighSoftWare96: Yep... Apparently it is a bug on their side..

HighSoftWare96:
> Just a suggest, will you consider replacing the whole block of `<script></script>` in `index.html`? I think you already changed all the content inside it.
> It is more simple to use a single replace.

Yeah, can do :)

giovanni-bertoncelli:
> Just a suggest, will you consider replacing the whole block of `<script></script>` in `index.html`? I think you already changed all the content inside it.
> It is more simple to use a single replace.

Yeah, that can do 😄

package.json
giovanni-bertoncelli: Yeah, probably a bug on their side

climba03003: @giovanni-bertoncelli
I finally see the cause, it should be the problem with bash, npm or yarn.
Can you test if adding a quote help to resolve it? It is a little bit too long if we include all the nested folder case.
```
"test": "standard && tap --100 'test/**/*.js' && npm run typescript",
"test:ci": "standard && tap 'test/**/*.js' --coverage-report=lcovonly && npm run typescript",
```

mcollina:
@giovanni-bertoncelli so we are waiting for a bit more change.

giovanni-bertoncelli:
> @giovanni-bertoncelli so we are waiting for a bit more change.

Yes. I'll replace the <script> and the package json in the next days..

package.json
giovanni-bertoncelli: Should be fine now

giovanni-bertoncelli:
Ok. I've:
- Fixed the package.json `npm run test` with quotes (as @climba03003 suggested)
- Replacing entirely the `<script>` tag (as @climba03003 suggested)
- Renamed the unit test 


_(Sorry for the double github account, they are both mine)_

lib/mode/dynamic.js
```diff
@@ -16,7 +16,8 @@ module.exports = function (fastify, opts, done) {
 
   if (opts.exposeRoute === true) {
     const prefix = opts.routePrefix || '/documentation'
-    fastify.register(require('../routes'), { prefix })
+    const uiConfig = opts.uiConfig || {}
```
climba03003, line 19:
```suggestion
    const uiConfig = Object.assign({}, opts.uiConfig)
```

climba03003 (approved):

mcollina (approved):
lgtm

giovanni-bertoncelli:
Any chance that this will be released soon? :)

Eomm:
Releasing 👍

giovanni-bertoncelli:
Sorry, I've seen the release but downloading with `npm` the resulting `static/` folder contains the old content... 🤔

Eomm:
mmm the script was updated
https://unpkg.com/browse/fastify-swagger@4.1.0/lib/util/prepare-swagger-ui.js

and it seems the pre-publish did not run:

https://github.com/fastify/fastify-swagger/blob/9d87a5fdfdcf4d7ffda39d01d4f696d2354bbf12/package.json#L7

I can release a fix this evening CEST if someone can't do it earlier

giovanni-bertoncelli:
Ok @Eomm is there any error in the package.json definition then?

mcollina:
I've released v4.1.1, try that.

giovanni-bertoncelli:
Yeah, now it's fine 🤷🏻‍♂️
Thanks @mcollina

PR #354, opened 2021-02-05
Default pkg to empty object in case readPackageJson fails

agramian:
This change addresses a regression when bundling with webpack after upgrading from v3 to v4 of this plugin.
The plugin fails after bundling because the `package.json` of this library is not present in the expected relative location and an attempt is made to read from an `undefined` object.

#### Checklist

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -35,8 +35,8 @@ function prepareDefaultOptions (opts) {
   }
 }
 
-function prepareOpenapiObject (opts, done) {
-  const pkg = readPackageJson(done)
+function prepareOpenapiObject (opts) {
+  const pkg = readPackageJson()
   const openapiObject = {
     openapi: '3.0.3',
     info: {
--- a/lib/spec/swagger/utils.js
+++ b/lib/spec/swagger/utils.js
@@ -46,7 +46,7 @@ function prepareDefaultOptions (opts) {
 }
 
 function prepareSwaggerObject (opts, done) {
-  const pkg = readPackageJson(done)
+  const pkg = readPackageJson(done) || {}
   const swaggerObject = {
     swagger: '2.0',
     info: {
--- a/lib/spec/swagger/utils.js
+++ b/lib/spec/swagger/utils.js
@@ -45,8 +45,8 @@ function prepareDefaultOptions (opts) {
   }
 }
 
-function prepareSwaggerObject (opts, done) {
-  const pkg = readPackageJson(done) || {}
+function prepareSwaggerObject (opts) {
+  const pkg = readPackageJson()
   const swaggerObject = {
     swagger: '2.0',
     info: {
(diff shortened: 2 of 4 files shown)


climba03003:
It will throw an error if `package.json` is not exist or not a valid `json`. I do not see a reason it is needed.
https://github.com/fastify/fastify-swagger/blob/master/lib/util/common.js#L95-L101

climba03003:
It should be a bug, we do not terminate the code execution after the error throwing.

Code before refactor
```JS
try {
  pkg = JSON.parse(fs.readFileSync(path.join(__dirname, 'package.json')))
} catch (err) {
  return next(err) // it will terminate the code execution
}
```

Code after refactor
```JS
const pkg = readPackageJson(done) // the code execution is not terminated
```

agramian:
Please see https://github.com/fastify/fastify/pull/2604 for a similar issue to this one in order to understand why this change is needed.

When bundling with webpack the location of the `package.json` being read at runtime will not be as expected by the relative path.

Reading the internal `package.json` file of this library should not be necessary to use it so if it fails it should not cause the entire program to terminate.
Instead it should just use a default value since this reading of the `pkg.version` and `pkg.name` is not essential functionality for this library.

This was working with webpack using v3 of this plugin but now this functionality has regressed with v4 due to this.

climba03003:
I just dont know why you can have `undefined` for pkg.
It will throw when executing `fs.readFileSync`, then it return `done(err)` for `pkg`.
`done` should be a `function`, so, it can never get `undefined` for `pkg`.

climba03003:
When I install `fastify-swagger@3.5.0`, missing `package.json` will result in an empty `API`
image
It only didn't block the startup. But it will end the prepare job will the file is missing.

climba03003 (changes requested):
Behavior for `3.5.0` - empty API definition when missing `package.json`
- Remove the try-catch in `readPackageJson`
- wrap it `prepareSwaggerObject` with try-catch
```JS
let swaggerObject
try {
  swaggerObject = prepareSwaggerObject(defOpts)
} catch(err) {
  return done(err)
}
```

Introduce new behavior - have API definition even without `package.json`
- update `readPackageJSON`, `return {}` when catch err
- remove all the necessary `done`

I prefer the latter.

agramian:
@climba03003 I made the changes you requested as I understood them however I did not add a test case.
Without the ability to mock functions or files I am unaware of a useful test case to make for this.
With `jest` I could mock the `readPackageJson` function but I could not find a way to mock functions using `tap`.
If you have a particular test case in mind let me know and I will add it.

climba03003:
For the test case, you can take this as an example.
https://github.com/fastify/fastify-swagger/blob/master/test/mode/static.js#L599-L632

1. change `test-package.json` to something invalid.
2. add `Swagger.validate` to validate it return a valid json schema.
3. change the text inside `t.pass`

climba03003 (commented):
For the test case, you can take this as an example.
https://github.com/fastify/fastify-swagger/blob/master/test/mode/static.js#L599-L632

1. change `test-package.json` to something invalid.
2. add `Swagger.validate` to validate it return a valid json schema.
3. change the text inside `t.pass`
4. change the `options` according to `openapi` dynamic mode and `swagger` dynamic mode

climba03003 (approved):
LGTM

mcollina (approved):
lgtm

PR #355, opened 2021-02-05
fix: missing security field in openapi

### LINKED ISSUE #352: OA3: global `security` configuration gets ignored
## 🐛 Bug Report

The global `security` configuration from OA3  (Open API 3) specification doesn't get exposed in the final JSON swagger configuration.
I'm trying to define the `security` configuration globally in the plugin settings like it is shown in the first example of this page: https://swagger.io/docs/specification/authentication/bearer-authentication/. It appears to me that this field it's ignored in the final result.

## To Reproduce

```js
  await fastify.register(
    require('fastify-swagger'),
    {
      exposeRoute: true,
      openapi: {
        info: {
          // infos
        },
        servers: [
          // servers
        ],
        components: {
          securitySchemes: {
            bearerAuth: {
              type: 'http',
              scheme: 'bearer',
              bearerFormat: 'JWT'
            }
          }
        },
        // this field gets ignored :(
        security: [{
          bearerAuth: []
        }]
      }
    }
  );
```

Then call `fastify.swagger()`: the result does not have a `security` key.

## Expected behavior

The result should export the `security` key.

## Workaround

As a workaround I can define a local `security` option for each exposed route by using the `transform` function.

## Hints

I debugged the issue and it seems that, in this point, the field is not expected:

https://github.com/fastify/fastify-swagger/blob/786567810498de5e6b488edd7cd1c22019be6289/lib/spec/openapi/utils.js#L23

I haven't understand if this plugin is completely compatible with OA3 so maybe this is still work in progress...

## Your Environment

- *node version*: 14.15.1
- *fastify version*: >=2.0.0
- *os*: Mac

climba03003:
Closes #352 

#### Checklist

### DIFF
--- a/examples/options.js
+++ b/examples/options.js
@@ -44,6 +44,18 @@ const openapiOption = {
     tags: [
       { name: 'tag' }
     ],
+    components: {
+      securitySchemes: {
+        apiKey: {
+          type: 'apiKey',
+          name: 'apiKey',
+          in: 'header'
+        }
+      }
+    },
+    security: [{
+      apiKey: []
+    }],
     externalDocs: {
       description: 'Find more info here',
       url: 'https://swagger.io'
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

PR #389, opened 2021-03-25
Support for complex serialization (eg. JSON) in query and cookie

sheldhur:
https://swagger.io/docs/specification/describing-parameters/#schema-vs-content

#### Checklist

- [x] run npm run test and npm run benchmark
- [x] commit message and code follows the Developer's Certification of Origin
and the Code of conduct


lib/util/common.js
```diff
@@ -68,8 +68,14 @@ function resolveLocalRef (jsonSchema, externalSchemas) {
   if (typeof jsonSchema.type !== 'undefined' && typeof jsonSchema.properties !== 'undefined') {
     // for the shorthand querystring/params/headers declaration
     const propertiesMap = Object.keys(jsonSchema.properties).reduce((acc, headers) => {
-      const required = (Array.isArray(jsonSchema.required) && jsonSchema.required.indexOf(headers) >= 0) || false
-      const newProps = Object.assign({}, jsonSchema.properties[headers], { required })
+      const rewriteProps = {}
```
climba03003, line 71: `common.js` is used in both `openapi` and `swagger`. It may cause problem when people mixing both behavior.

climba03003 (commented):
I assume the same syntax of config should be available for both `openapi` and `swagger`. It need to be supported in both standard.

Documentation is missing.

sheldhur:
@climba03003 I can't found anything about this in documentation for 2.0 version. Only for 3.0+

lib/util/common.js
sheldhur: All tests passed.

climba03003: Yes, it is. but what I means is the same passed variables for `swagger`. It should throw an error when validating because it do not handle the `rawRequired`, etc. It should be a missing test case.

sheldhur: Got it. I'll think something

sheldhur: @climba03003 if I throw an exception, the user will not see it without logger. I have not found any exceptions in the project. Do you really want throw an exception? Or I can just delete `rawRequired` and `consume` fields in swagger

climba03003: It should be visible when calling `fastify.swagger`, if they follow the document.
```
fastify.ready(err => {
  if (err) throw err
  fastify.swagger() // it throw in here
})
```
or they cannot get a proper `json` when access `/documentation/json`.

test/spec/openapi/route.js
```diff
@@ -419,3 +419,105 @@ test('cookie, query, path description', t => {
       })
   })
 })
+
+test('cookie and query with serialization type', t => {
```
mcollina, line 423: please use async/await in this test

mcollina (changes requested):

test/spec/openapi/route.js
sheldhur: ready

lib/util/common.js
sheldhur: Ready. `plainJsonObjectToSwagger2` not throw exception

climba03003 (approved):
LGTM. and I would like to hear feedback from @Eomm or @SkeLLLa before merging.

test/spec/openapi/route.js
climba03003: I think the `async / await` should be moved up one more level, like `async (t) => {}`

Eomm:
I will check it later these days 👍🏼

mcollina (approved):
lgtm

sheldhur:
@Eomm 👀

lib/spec/openapi/utils.js
```diff
@@ -99,8 +99,24 @@ function plainJsonObjectToOpenapi3 (container, jsonSchema, externalSchemas) {
         const result = {
           in: container,
           name: propertyName,
-          required: jsonSchemaElement.required,
-          schema: jsonSchemaElement
+          required: jsonSchemaElement.required
+        }
+        // complex serialization in query or cookie, eg. JSON
+        // https://swagger.io/docs/specification/describing-parameters/#schema-vs-content
+        if (jsonSchemaElement.consume) {
```
Eomm, line 106: as other PRs, it could be more friendly to adopt a name like `x-consume` to warn the uses that, that field is not standard

lib/spec/openapi/utils.js
```diff
@@ -99,8 +99,24 @@ function plainJsonObjectToOpenapi3 (container, jsonSchema, externalSchemas) {
...
+          result.content = {
+            [jsonSchemaElement.consume]: {
+              schema: {
+                ...jsonSchemaElement,
+                required: jsonSchemaElement.rawRequired
```
Eomm, line 111: `rawRequired` is a interna utility, we could use `Symbols` to hide it from the user

lib/spec/openapi/utils.js
sheldhur: Ready. "x-consume" with "x-response-description" moved to constants

lib/spec/openapi/utils.js
sheldhur: ready

Eomm (approved):

PR #393, opened 2021-04-02
feat: move to tap 15.0.1

zekth:


mcollina (approved):
lgtm

zekth:
FYI: had to add the snapshot file in the scm.

mcollina:
@zekth you need to remove https://github.com/fastify/fastify-swagger/blob/master/tap-snapshots/test-mode-static.js-TAP.test.js

Issue #403, opened 2021-04-15
feat: add option to hide untagged routes


DRoet:
## 🚀 Feature Proposal

Instead of adding `hide: true` to every route we want to exclude, we could make an option to hide everything that is untagged.  All the routes that we actually want to include are usually already tagged anyway.

## Motivation

One of the useful features that [fastify-oas](https://github.com/SkeLLLa/fastify-oas#plugin-options) implemented.

## Example
```js
fastify.register(require('fastify-swagger'), {
   openapi: {...},
   hideUntagged: true,
})
```

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

zekth:
couldn't you use `hiddenTag` option instead? see: https://github.com/fastify/fastify-swagger/pull/278

DRoet:
I would prefer a global option to ignore everything that's not tagged, `hiddenTag` still requires tagging to opt-out... I would rather opt-in.

DRoet:
Thanks! happy to say I was able to migrate back to `fastify-swagger` 😄

PR #406, opened 2021-04-21
added option to hide untagged routes

KiraPC:
#### Checklist

- [x] run `npm run test` 

Added a new plugin parameter **hideUntagged**.
Added cases in both dynamic swagger and openapi example
Added tests cases in both dynamic swagger and openapi
Update readme with the new parameter description

Related issue: https://github.com/fastify/fastify-swagger/issues/403

### DIFF
--- a/lib/util/common.js
+++ b/lib/util/common.js
@@ -41,7 +41,9 @@ function addHook (fastify) {
   }
 }
 
-function shouldRouteHide (schema, hiddenTag, hideUntagged) {
+function shouldRouteHide (schema, opts) {
+  const { hiddenTag, hideUntagged } = opts
+
   if (schema && schema.hide) {
     return true
   }
@@ -52,9 +54,10 @@ function shouldRouteHide (schema, hiddenTag, hideUntagged) {
     return true
   }
 
-  if (schema && schema.tags && schema.tags.includes(hiddenTag)) {
+  if (tags.includes(hiddenTag)) {
     return schema.tags.includes(hiddenTag)
   }
+
   return false
 }
 
--- a/lib/util/common.js
+++ b/lib/util/common.js
@@ -41,10 +41,17 @@ function addHook (fastify) {
   }
 }
 
-function shouldRouteHide (schema, hiddenTag) {
+function shouldRouteHide (schema, hiddenTag, hideUntagged) {
   if (schema && schema.hide) {
     return true
   }
+
+  const tags = (schema && schema.tags) || []
+
+  if (tags.length === 0 && hideUntagged) {
+    return true
+  }
+
   if (schema && schema.tags && schema.tags.includes(hiddenTag)) {
     return schema.tags.includes(hiddenTag)
   }
--- a/test/spec/swagger/option.js
+++ b/test/spec/swagger/option.js
@@ -386,6 +386,39 @@ test('hide support - tags Custom', t => {
   })
 })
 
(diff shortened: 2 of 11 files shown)


lib/util/common.js
```diff
@@ -41,10 +41,17 @@ function addHook (fastify) {
   }
 }
 
-function shouldRouteHide (schema, hiddenTag) {
+function shouldRouteHide (schema, hiddenTag, hideUntagged) {
```
zekth, line 44: I'd rather pass the option object rather than adding another parameter. 3 parameters is still acceptable in conventions but as those 2 last parameters belongs from the same object; to me it would be better to factorize

zekth (changes requested):
Minor refactor to me. Maybe @climba03003 @Eomm have another opinion.

lib/util/common.js
KiraPC: I got your point. No problem to factorize.

Do you prefer to pass the entire opts object or create a new one that includes just the shouldRouteHide options?

In the feature if the shouldRouteHide need to accept more parameters we can use that options object instead modify the plugin one.

lib/util/common.js
```diff
@@ -41,10 +41,17 @@ function addHook (fastify) {
...
     return true
   }
+
+  const tags = (schema && schema.tags) || []
+
+  if (tags.length === 0 && hideUntagged) {
+    return true
+  }
+
   if (schema && schema.tags && schema.tags.includes(hiddenTag)) {
```
climba03003, line 55: You have ensure the tags be an array in above of code. No need to check it like this in here. And `tags.includes(hiddenTag)` is duplicate work.
```suggestion
  if (tags.includes(hiddenTag)) {
    return true
```

climba03003 (changes requested):

lib/util/common.js
KiraPC: @climba03003 
how careless I am? I forgot. I will add this in PR after feedback on my question above.

climba03003: It is not your fault, it is the code in the pass. But with your changes, it be better to update together.

lib/util/common.js
KiraPC: @zekth I committed with my preferred solution. I used a new option object to avoid passing the main plugin option.

If it doesn't match your requirements, I'll update.

zekth (approved):
lgtm. Thanks!

climba03003 (approved):
LGTM.

test/spec/swagger/option.js
```diff
@@ -386,6 +386,39 @@ test('hide support - tags Custom', t => {
...
+      }
+    }
+  }
+
+  fastify.get('/', opts, () => {})
+
+  fastify.ready(err => {
+    t.error(err)
+
+    const openapiObject = fastify.swagger()
```
climba03003, line 417:
```suggestion
    const swaggerObject = fastify.swagger()
    t.notOk(swaggerObject.paths['/'])
```

climba03003 (changes requested):
Missed this one. Use the right naming for the test case.

test/spec/swagger/option.js
KiraPC: Damn, the copy and paste :)

Done!

climba03003 (approved):
LGTM. Thanks for your works.

PR #427, opened 2021-06-20
Avoid inline script execution

Dschoordsch:
Chrome refuses to execute inline scripts when certain headers are set.
This is simple to avoid by moving the init script into a separate file.

#### Checklist

### DIFF
--- a/lib/util/prepare-swagger-ui.js
+++ b/lib/util/prepare-swagger-ui.js
@@ -24,8 +24,10 @@ filesToCopy.forEach(filename => {
 })
 
 const newIndex = fs.readFileSync(resolve('./static/index.html'), 'utf8')
-  .replace(/<script>((.| |\n)*)<\/script>/, `
-  <script>
+  .replace(/<script>((.| |\n)*)<\/script>/, '<script src="./swagger-ui-init.js" charset="UTF-8"> </script>')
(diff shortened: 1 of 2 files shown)


climba03003:
I think you can use the `staticCSP` option to prevent the block of script execution.

Dschoordsch:
> I think you can use the `staticCSP` option to prevent the block of script execution.

You're probably right. I assume I can also change or disable the security headers for this route. I stumbled across this issue because the UI stopped working after working on unrelated parts, the security headers were probably set by another plugin.
I don't see a downside in putting the init script into a separate file, I think this gives a better out-of-the-box experience because the user doesn't accidentally stumble into this situation and then has to figure out what's happening. Maybe I'm missing something?

climba03003:
Although, `script` is the critical part of `swagger-ui` that can be run. It will still block the image and css if the CSP header configuration is wrong. Ideally, I want it to be solved for all resources instead of script only. Then, we can remove the CSP part of code from this project.

sergeycrisp (approved):

mcollina (approved):
lgtm

mcollina:
Oh no, apparently we never landed this. This now conflicts, would you like to rebase? I'll land when things pass.

Fdawgs:
Will be resolved by #536

Issue #337, opened 2021-01-26
References to "definitions" 


jsumners:
## 🐛 Bug Report

The "models" builder does not correctly reference embedded "definitions" style references.

## To Reproduce

**package.json**
```js
{
  "dependencies": {
    "fastify": "^3.11.0",
    "fastify-swagger": "^3.5.0"
  }
}
```

**index.mjs**
```js
"use strict";

import Fastify from "fastify";
import * as fastifySwagger from "fastify-swagger";

const fastify = Fastify();
fastify.register(fastifySwagger, {
  routePrefix: "/_swagger",
  exposeRoute: true,
  swagger: {
    info: {
      title: "Test Swagger",
    },
  },
});

fastify
  .addSchema({
    $id: "urn:schema:base",
    definitions: {
      hello: { type: "string" },
    },
    type: "object",
    properties: {
      hello: { $ref: "#/definitions/hello" },
    },
  })
  .addSchema({
    $id: "urn:schema:ref",
    type: "object",
    properties: {
      hello: { $ref: "urn:schema:base#/definitions/hello" },
    },
  });

fastify.route({
  path: "/",
  method: "GET",
  schema: {
    response: {
      "2xx": fastify.getSchema("urn:schema:ref"),
    },
  },
  handler(req, res) {
    res.send({ hello: "world" });
  },
});

const response = await fastify.inject({ url: "/_swagger/json" });
console.log(JSON.stringify(response.json(), null, 2));
```

## Expected behavior

The above reproduction will produce the following JSON:

```json
{
  "swagger": "2.0",
  "info": {
    "title": "Test Swagger"
  },
  "definitions": {
    "def-0": {
      "definitions": {
        "hello": {
          "type": "string"
        }
      },
      "type": "object",
      "properties": {
        "hello": {
          "$ref": "#/definitions/hello"
        }
      }
    },
    "def-1": {
      "type": "object",
      "properties": {
        "hello": {
          "$ref": "urn:schema:base#/definitions/hello"
        }
      }
    }
  },
  "paths": {
    "/": {
      "get": {
        "responses": {
          "2xx": {
            "schema": {
              "$id": "todo.com",
              "type": "object",
              "properties": {
                "hello": {
                  "type": "string"
                }
              }
            },
            "description": "Default Response"
          }
        }
      }
    }
  }
}
```

[...]

PR #432, opened 2021-06-29
Add support for $ref in header schemas

cq-fpietron:
Greetings!

First of all, thanks for all the work on the fastify ecosystem. It's a blast to utilise it in projects.

This PR includes a small adjustment to the way `header`s are treated in `plainJsonObjectToOpenapi3`.

The motivation is a scenario from my setup, where I attach some headers validation to most of my routes. They contain either `enum`s or `example`s that I'd like to be properly passed down to the resulting OpenAPI schema. One of the benefits of this is, in tandem with properly initialising refs, to have its presentation in `swagger-ui` improved: select fields for `enum` etc.

I first experimented with injecting `example` and `examples` fields directly into the value returned from `toOpenapiProp`, however I think the approach with `$ref` is both more elegant and functional.

Please let me know whether: my suggestion is too narrow or not, and is the added test sufficient enough.

 I'm also wondering 
 - if it shouldn't search for `$ref` within an additional `schema` property,
 - whether the `type` field next to possible `$ref` should now be optional or not - best case scenario I'd like to define the `$ref` only; same goes for `description`?

PS: isn't the PR template outdated? I see no `benchmark` script in `package.json`.

Cheers!

#### Checklist

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -158,7 +158,7 @@ function plainJsonObjectToOpenapi3 (container, jsonSchema, externalSchemas) {
       break
     case 'header':
       toOpenapiProp = function (propertyName, jsonSchemaElement) {
-        return {
+        const result = {
           in: 'header',
           name: propertyName,
           required: jsonSchemaElement.required,
@@ -167,6 +167,9 @@ function plainJsonObjectToOpenapi3 (container, jsonSchema, externalSchemas) {
             type: jsonSchemaElement.type
           }
         }
+
+        if (jsonSchemaElement.$ref) result.schema.$ref = jsonSchemaElement.$ref
+        return result
       }
       break
   }
--- a/test/spec/openapi/option.js
+++ b/test/spec/openapi/option.js
@@ -607,4 +607,55 @@ test('uses examples if has multiple array examples', t => {
   })
 })
 
+test('passes $ref to header schema if specified', t => {
+  t.plan(6)
+  const headerToReference = 'someHeader'
+
+  const fastify = Fastify()
+  fastify.addSchema({ $id: `#/components/headers/${headerToReference}`, type: 'string' })
+
+  fastify.register(fastifySwagger, openapiOption)
+
+  const opts = {
+    schema: {
+      headers: {
+        type: 'object',
+        required: ['referencedHeader'],
+        properties: {
+          referencedHeader: {
+            $ref: `#/components/headers/${headerToReference}`
+          }
+        }
+      },
+      body: {
+        type: 'object',
+        required: ['hello'],
+        properties: {
+          hello: {
+            type: 'string'
+          }
+        }
+      }
+    }
+  }
+
+  fastify.get('/', opts, () => {})
+
+  fastify.ready(err => {
+    t.error(err)
+
+    const openapiObject = fastify.swagger()
+    const parameters = openapiObject.paths['/'].get.parameters
+
+    t.ok(parameters)
+    t.equal(parameters.length, 1)
+
+    const headerParameter = parameters[0]
+
+    t.ok(headerParameter)
+    t.ok(headerParameter.schema)
+    t.equal(headerParameter.schema.$ref, opts.schema.headers.properties.referencedHeader.$ref)
+  })
+})
+
 module.exports = { openapiOption }


lib/spec/openapi/utils.js
```diff
@@ -167,6 +167,9 @@ function plainJsonObjectToOpenapi3 (container, jsonSchema, externalSchemas) {
             type: jsonSchemaElement.type
           }
         }
+
+        if (jsonSchemaElement.$ref) result.schema.$ref = jsonSchemaElement.$ref
```
cq-fpietron, line 171: I'm wondering whether this should be
```ts
if (jsonSchemaElement.$ref) {
   result.schema.$ref = jsonSchemaElement.$ref
   delete result.schema.type
   delete result.description
 }
```

To avoid `undefined` fields in `fastify.swagger()`?

mcollina (approved):
lgtm

lib/spec/openapi/utils.js
zekth: As you pass a `$ref` the property `type` has no purpose to exist, on the other hand `description` is still legit to be here. So technically you should only `delete result.schema.type`

climba03003:
After checking the generated json, I think this cannot be supported unless we add the support of object property `$ref` resolution.
The output and the test seems to be fine, but you can see that `components.schemas` is empty json. `swagger-ui` will throw error because `#/components/headers/someHeader` cannot be find.

```js
{
  "openapi": "3.0.3",
  "info": {
    "title": "Test swagger",
    "description": "testing the fastify swagger api",
    "version": "0.1.0"
  },
  "components": {
    "securitySchemes": {
      "apiKey": {
        "type": "apiKey",
        "name": "apiKey",
        "in": "header"
      }
    },
    "schemas": {}
  },
  "paths": {
    "/": {
      "get": {
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "required": [
                  "hello"
                ],
                "properties": {
                  "hello": {
                    "type": "string"
                  }
                }
              }
            }
          }
        },
        "parameters": [
          {
            "in": "header",
            "name": "referencedHeader",
            "required": true,
            "schema": {
              "$ref": "#/components/headers/someHeader"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "Default Response"
          }
        }
      }
    }
  },
  "servers": [
    {
      "url": "http://localhost"
    }
  ],
  "security": [
    {
      "apiKey": []
    }
  ],
  "tags": [
    {
      "name": "tag"
    }
  ],
  "externalDocs": {
    "description": "Find more info here",
    "url": "https://swagger.io"
  }
}
```

cq-fpietron:
@climba03003 thanks for your checks. I have stumbled into this too. I was able to get it to work in a slightly hacky way, though.

The solution was to use `$ref`s to `components/schemas` instead of `parameters` or `headers`(As per OpenAPI 3.0 [components section](https://swagger.io/docs/specification/components/), `components/headers` are for _response_ headers, not _request_ headers). I guess it is not as crucial in this case as I have only seen references to `components/schemas` in `fastify-swagger`'s codebase.

I also had those definitions added to fastify via `app.addSchema` with `$id: '#/components/schemas/someHeader`.

It would appear that parsing `components/parameters` is nearly valid, as I can either get fully valid OA3 schema or one that displays in swagger-ui nicely. I have created a sample repository available here to depict this behaviour: [cq-fpietron/fastify-swagger-refs](https://github.com/cq-fpietron/fastify-swagger-refs). 

Starting up the server you can toggle between registering `fastifySwagger` with either option.

@climba03003 , do you have any idea how we could estimate work on getting this feature right?

climba03003:
> 
> 
> @climba03003 thanks for your checks. I have stumbled into this too. I was able to get it to work in a slightly hacky way, though.
> 
> The solution was to use `$ref`s to `components/schemas` instead of `parameters` or `headers`(As per OpenAPI 3.0 [components section](https://swagger.io/docs/specification/components/), `components/headers` are for _response_ headers, not _request_ headers). I guess it is not as crucial in this case as I have only seen references to `components/schemas` in `fastify-swagger`'s codebase.
> 
> I also had those definitions added to fastify via `app.addSchema` with `$id: '#/components/schemas/someHeader`.
> 
> It would appear that parsing `components/parameters` is nearly valid, as I can either get fully valid OA3 schema or one that displays in swagger-ui nicely. I have created a sample repository available here to depict this behaviour: [cq-fpietron/fastify-swagger-refs](https://github.com/cq-fpietron/fastify-swagger-refs).
> 
> Starting up the server you can toggle between registering `fastifySwagger` with either option.
> 
> @climba03003 , do you have any idea how we could estimate work on getting this feature right?

The fastest way is update the code of resolver in https://github.com/Eomm/json-schema-resolver which fit the needs.
The best way is rewrite a brand new resolver which can solve https://github.com/fastify/fastify-swagger/issues/337

davidschmitt:
FYI, I have created a drop-in replacement for `json-schema-resolver` which solves many of the `$ref` related issues in external schemas for `fastify-swagger`.

I've published the code here:  https://github.com/davidschmitt/fastify-schema-resolver.git

I haven't published the package on `npmjs.com` (I've never done that before), but if the `fastify-swagger` maintainers are interested in switching I can do so.

PR #444, opened 2021-07-29
All

bkulyk:
Modify regex for finding path params to ignore path verbs (they start with `::`), and replace `::` with `:`  for final render in swagger.

> If you want a path containing a colon without declaring a parameter, use a double colon. For example:
> `fastify.post('/name::verb') // will be interpreted as /name:verb`

[Fastify docs # UrlBuilding](https://www.fastify.io/docs/latest/Routes/#url-building)

#### Checklist

- [x] run `npm run test` and `npm run benchmark` (benchmark command seems missing)
- [x] documentation is changed or added (think it's expected to work based off fastify docs)


mcollina (approved):
lgtm

PR #447, opened 2021-08-01
Add support for async transformers

### LINKED ISSUE #446: Async transformer
## 🚀 Feature Proposal

Add support for async functions/Promise returning functions to the `transform` option.

## Motivation

This would allow the use of libraries/methods of transforming schemas that are asynchronous for one reason or another. As an example, [json-schema-to-openapi-schema](https://github.com/openapi-contrib/json-schema-to-openapi-schema) is strictly async, thus it cannot be used with the current synchronous-only transform syntax.

## Example

```js
const convert = require('@openapi-contrib/json-schema-to-openapi-schema');

fastify.register(require('fastify-swagger'), {
  swagger: { ... },
  ...
  transform: async schema => {
    const {
      params = undefined,
      body = undefined,
      querystring = undefined,
      ...others
    } = schema
    const transformed = { ...others }
    if (params) transformed.params = await convert(params)
    if (body) transformed.body = await convert(body)
    if (querystring) transformed.querystring = await convert(querystring)
    return transformed
  }
}
```

---

I've tried doing this on a branch just before opening this issue, then thought it would be best to discuss this as I noticed that the way I implemented it—making the `swagger` function `async` and then changing the transform call to `await defOpts.transform(route.schema)`—would result in a pretty big breaking change as all `fastify.swagger()` calls would end up returning a Promise instead of just the schema. Perhaps instead it could only end up becoming a async function if the `transform` option is an async function as well, thus preserving existing code, or make it an option passed `fastify.swagger()`—`async: true` most likely (but then how do we handle an async transformer being called without this option being passed?)

Ovyerus:
Add support for providing an asynchronous function to `transform` in options, to allow using some libraries that need to be awaited.  When a Promise-returning/`async` function is provided to `transform`, `fastify.swagger()` returns a Promise instead of just the schema object, so this doesn't bring about a breaking change, as discussed in #446.

Closes #446.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,9 @@ If you set `exposeRoute` to `true` the plugin will expose the documentation with
 <a name="register.options.transform"></a>
 #### Transforms
 
-To use different schemas such as [Joi](https://github.com/hapijs/joi) you can pass a synchronous `transform` method in the options to convert them back to standard JSON schemas expected by this plugin to generate the documentation (`dynamic` mode only).
+To use different schemas such as [Joi](https://github.com/hapijs/joi) you can pass a synchronous or asynchronous `transform` method in the options to convert them back to standard JSON schemas expected by this plugin to generate the documentation (`dynamic` mode only).
+
+##### Synchronous transform
 
 ```js
 const convert = require('joi-to-json')
@@ -250,6 +252,30 @@ fastify.register(require('fastify-swagger'), {
     if (params) transformed.params = convert(params)
     if (body) transformed.body = convert(body)
     if (querystring) transformed.querystring = convert(querystring)
+    return transformed
+  }
+}
+```
+
+##### Asynchronous transform
+
+```js
+const convert = require('@openapi-contrib/json-schema-to-openapi-schema')
+
+fastify.register(require('fastify-swagger'), {
+  swagger: { ... },
+  ...
+  transform: async schema => {
+    const {
+      params = undefined,
+      body = undefined,
+      querystring = undefined,
+      ...others
+    } = schema
+    const transformed = { ...others }
+    if (params) transformed.params = await convert(params)
+    if (body) transformed.body = await convert(body)
+    if (querystring) transformed.querystring = await convert(querystring)
     return transformed
   }
 }
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ fastify.register(require('fastify-swagger'), {
 ```
 
 ##### Asynchronous transform
+**NOTE:** When using an async transformer, any calls of `fastify.swagger()` will need to be awaited (or chained  with `.then`) since it will return a Promise instead of just a JSON object.
 
 ```js
 const convert = require('@openapi-contrib/json-schema-to-openapi-schema')
--- a/test/transform.js
+++ b/test/transform.js
@@ -11,9 +11,11 @@ const params = Joi
   .keys({
     property: Joi.string().required()
   })
+
 const opts = {
   schema: { params }
 }
+
 const convertible = ['params', 'body', 'querystring']
 const valid = {
   transform: schema => Object.keys(schema).reduce((transformed, key) => {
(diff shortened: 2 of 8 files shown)


test/transform.js
```diff
@@ -55,3 +60,77 @@ test('transform should work with a Function', t => {
...
+    openapi: {},
+    transform: async schema => valid.transform(schema)
+  })
+
+  fastify.setValidatorCompiler(({ schema }) => Joi.validate(schema))
+  fastify.get('/example', opts, () => {})
+
+  fastify.ready(err => {
+    t.error(err)
+    t.resolves(fastify.swagger)
```
climba03003, line 78: You need to check if the output is valid or not.

Ovyerus: Done 👍

README.md
```diff
@@ -255,6 +257,30 @@ fastify.register(require('fastify-swagger'), {
 }
 ```
 
+##### Asynchronous transform
```
climba03003, line 260: Can you mention about if you use `async transform`, you need to await `fastify.swagger`. Otherwise, it will return `Promise` and may break your application.

climba03003 (changes requested):
Routes also need some change, currently the registered route only support `sync` return.

https://github.com/fastify/fastify-swagger/blob/master/lib/routes.js#L86
https://github.com/fastify/fastify-swagger/blob/master/lib/routes.js#L97

mcollina (commented):
I would note that it's possible to _not_ change the public API, i.e. keep `fastify.swagger()` synchronous.

If you look into https://github.com/fastify/fastify-swagger/blob/master/lib/mode/dynamic.js#L5-L36, you can see we can change this function to `async` like so:

```js
module.exports = async function (fastify, opts) {
  const { routes, Ref } = addHook(fastify)

  opts = Object.assign({}, {
    exposeRoute: false,
    hiddenTag: 'X-HIDDEN',
    hideUntagged: false,
    stripBasePath: true,
    openapi: null,
    swagger: {},
    transform: null
  }, opts)

    const uiConfig = opts.uiConfig || {}
    const initOAuth = opts.initOAuth || {}
    const staticCSP = opts.staticCSP
    const transformStaticCSP = opts.transformStaticCSP
    fastify.register(require('../routes'), { prefix, uiConfig, initOAuth, staticCSP, transformStaticCSP })
  }

  const cache = {
    object: null,
    string: null
  }
 
 fastify.addHook('onReady', async function () {
  
  const swagger = await resolveSwaggerFunction(opts, cache, routes, Ref)
  fastify.swagger = swagger
 })

  fastify.decorate('swagger', null)
}
```

As you can see, the `cache` in https://github.com/fastify/fastify-swagger/blob/095e8fc33949c93dbcfca493218c914210786c22/lib/spec/openapi/index.js#L13-L17 already makes this method synchronous. Therefore we can move the preparation of the swagger to the outer function and always return the cached value.

climba03003 (commented):
@Ovyerus Can you take a look on mcollina advice?

Ovyerus:
@mcollina While the solution wouldn't be exactly that, I guess we could do a sort of pre-compute by running `fastify.swagger()` in the ready function as specified, so that any future calls hit the cache. I'll give it a shot and see how it turns out.

mcollina:
> @mcollina While the solution wouldn't be exactly that, I guess we could do a sort of pre-compute by running `fastify.swagger()` in the ready function as specified, so that any future calls hit the cache. I'll give it a shot and see how it turns out.

+1 to that.

Ovyerus:
For some reason this change is making `test/spec/openapi/refs` fail, `ENOENT: no such file or directory, open '...\fastify-swagger\Order'`, but I'm not seeing any particular reason why it's failing 🤔

Ovyerus:
Long time no see; been busy with work so this fell out of mind unfortunately.

Seems this pre-cache method (calling `fastify` in an `onReady` hook) is still screwing with some of the tests, even moreso now actually. Seems to only be a bunch of `ref` related ones. I have no idea why, one of my friends suspects a race condition but how to find that - if there even is one - is beyond me. (maybe it needs some time to get ready for some reason?)

Personally I preferred the previous method of needing to call `fastify.swagger` with await manually, and it worked much better than this.

Issue #286, opened 2020-08-19
Schema added using fastify.addSchema is resulted as "def-#" in Swagger Models section


velopert:
## 🐛 Bug Report

When I add schema by `fastify.addSchema`, the Model name is shown as `def-#` where `#` is index.

![image](https://user-images.githubusercontent.com/17202261/90682552-1567f280-e2a0-11ea-9897-6f7b85eb2130.png)

## To Reproduce

```js
const fastify = require('fastify')();

fastify.register(require('fastify-swagger'), {
  swagger: {
    info: {
      title: 'Test swagger',
      description: 'testing the fastify swagger api',
      version: '0.1.0',
    },
    host: 'localhost',
    schemes: ['http'],
    consumes: ['application/json'],
    produces: ['application/json'],
  },
  exposeRoute: true,
});

fastify.addSchema({
  $id: 'greetings',
  type: 'object',
  properties: {
    hello: { type: 'string' },
  },
});

fastify.put(
  '/some-route/:id',
  {
    schema: {
      description: 'post some data',
      tags: ['user'],
      summary: 'qwerty',
      params: {
        type: 'object',
        properties: {
          id: {
            type: 'string',
            description: 'user id',
          },
        },
      },
      response: {
        201: {
          description: 'Succesful response',
          type: 'object',
          properties: {
            greetings: { $ref: 'greetings#' },
          },
        },
      },
    },
  },
  (req, reply) => {}
);

fastify.listen(5000, (err) => {
  if (err) throw err;
  console.log('listening');
});

```

## Expected behavior

The model names should be shown properly like following screenshot:

![image](https://user-images.githubusercontent.com/17202261/90682804-80b1c480-e2a0-11ea-9626-9b1cf9661f60.png)

## Your Environment

- *node version*: v12.15.0
- *fastify version*: ^3.2.1
- *os*: Mac
- "fastify-swagger": "^3.3.0"
- *any other relevant information*

fox1t:
Would you mind to send a PR to address this issue?

velopert:
I tried to fix the issue, however, my solution breaks the test code 😂.
I will send a PR when I get to fix the issue properly. 

Thanks!

pierlucg-xs:
@velopert Are you making progress? I would be interested in this and I'd be willing to give a hand.

pierlucg-xs:
Upon further looking, I saw that this is caused by https://github.com/Eomm/json-schema-resolver (the example in the readme demonstrates this)

[...]

Issue #225, opened 2020-02-13
Swagger Docs generated do not show added schemas as models


wolfejw86:
# Feature Request

While fastify-swagger is great for schemas attached to routes, I also would expect to be able to leverage it for displaying the models independent of the routes (usually shown at the bottom of the generated swagger-ui).  When I call `fastify.addSchema` on one of my reusable schemas, I would expect it to show up as a model.

## To Reproduce

1. generate a boilerplate fastify app with `fastify-cli`

Paste your code here:
2. change the `app.js` file to the following (basically copy / pasted from fastify-swagger + my example of trying to use the schema to display a model
```js
"use strict";

const path = require("path");
const AutoLoad = require("fastify-autoload");

module.exports = function(fastify, opts, next) {
  // Place here your custom code!

  // Do not touch the following lines

  // This loads all plugins defined in plugins
  // those should be support plugins that are reused
  // through your application
  fastify.register(AutoLoad, {
    dir: path.join(__dirname, "plugins"),
    options: Object.assign({}, opts)
  });

  // This loads all plugins defined in services
  // define your routes in one of these
  fastify.register(AutoLoad, {
    dir: path.join(__dirname, "services"),
    options: Object.assign({}, opts)
  });

  fastify.register(require("fastify-swagger"), {
    routePrefix: "/documentation",
    swagger: {
      info: {
        title: "Test swagger",
        description: "testing the fastify swagger api",
        version: "0.1.0"
      },
      externalDocs: {
        url: "https://swagger.io",
        description: "Find more info here"
      },
      host: "localhost",
      schemes: ["http"],
      consumes: ["application/json"],
      produces: ["application/json"],
      tags: [
        { name: "user", description: "User related end-points" },
        { name: "code", description: "Code related end-points" }
      ],
      securityDefinitions: {
        apiKey: {
          type: "apiKey",
          name: "apiKey",
          in: "header"
        }
      }
    },
    exposeRoute: true
  });

  const schemaIWantToAdd = {
    $id: "User",
    type: "object",
    required: ["id", "firstName", "lastName", "email"],
    properties: {
      id: { type: "string", format: "uuid" },
      firstName: { type: "string" },
      lastName: { type: "string" },
      email: { type: "string", format: "email" }
    }
  };

  // trying to use the schema as a model below
  fastify.addSchema(schemaIWantToAdd);

  fastify.ready(err => {
    if (err) throw err;
    fastify.swagger();
  });

  fastify.get(
[...]

PR #459, opened 2021-08-30
feat: customize def-0 model name

Eomm:
This PR:

- fixes the **swagger UI layout** to show the schema's `$id` as model name (see screenshot at the bottom). Under the hood, the `def-0` renaming is still in place.
- adds documentation to describe what is happening under the hood.
- adds the option to customize the `$ref`'s link into the inline schema. This gives to the user the chance to do what he/she wants and remove totally the `def-0` logic.

I think this (too simple?) solution solves the case.
Feel free to give feedback about this PR

### Why this change

To avoid a major bump, this PR fixes the view issue without breaking the `def-0` logic that has been developed to protect users actually 😄 

Let me explain.

This JSON Schema has a relative URI `$id`. 

```
fastify.addSchema({
  $id: 'user',
  type: 'object',
  definitions: {
    id: {
      type: 'string',
      description: 'user id'
    }
  }
})
```

This is wrong from the [DRAFT 07 specification](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01#section-8.2.1):

>    The root schema of a JSON Schema document SHOULD contain an "$id"
>   keyword with an absolute-URI [RFC3986] (containing a scheme, but no
>   fragment), or this absolute URI but with an empty fragment.

Moreover, referencing the `user` schema with the `$ref: 'user#/definitions/id'` is wrong from the specification point of view: the reference is relative and it should be resolved using the local schema's root URI, not the external one (the `$id: 'user'` one to be clear).

The `fastify-swagger` knows that the `$id` field is used as a simple key/value pair actually.
For this reason, it ignores the user's `$id`*** and deference the schema to a `def-0`: to be sure to avoid conflicts to let the `swagger-ui` module resolve correctly the references. Otherwise, you may have different output based on the `$id`s the developer set.
The specification, cover a lot more use cases than I saw in the GitHub issues actually.

The last example, setting an absolute root URI tells to the `swagger-ui` module to fetch the schema from the link!

```
fastify.addSchema({
  $id: 'http://example.com/user.json',
  ...
```

This snippet causes the swagger-ui module to make an HTTP request to `http://example.com/user.json`!

Now, forcing `fastify-swagger` to use the `$id` can break the configuration of those devs that use the standard strictly.
Using the `$id` in place of `def-0` is correct only if the developer knows that there is no conflicts or nested schema's declarations. 
With this PR, the user may replace the `buildLocalReference` option to accomplish this. But this behaviour leads to a "convention over configuration" strategy.

Before|After|
|---|---|
image | image

Fixes https://github.com/fastify/fastify-swagger/issues/337
Fixes https://github.com/fastify/fastify-swagger/issues/286
Fixes https://github.com/fastify/fastify-swagger/issues/225
Should help https://github.com/fastify/fastify-swagger/pull/432

\*** when I say that this module "ignores" the `$id`s I mean that we know that all the `$ref` can be resolved locally, accessing the data added to the fastify schema via the `addSchema` method. If the user set an absolute URI as `$id`, fastify will never look up an external link.

### DIFF
--- a/test/spec/openapi/schema.js
+++ b/test/spec/openapi/schema.js
@@ -308,5 +308,25 @@ test('support global schema reference', async t => {
 
   const swaggerObject = fastify.swagger()
   const api = await Swagger.validate(swaggerObject)
-  t.same(api.components.schemas['def-0'], schema)
+  t.match(api.components.schemas['def-0'], schema)
+})
+
+test('support global schema reference with title', async t => {
+  const schema = {
+    title: 'schema view title',
+    type: 'object',
+    properties: {
+      hello: { type: 'string' }
+    },
+    required: ['hello']
+  }
+  const fastify = Fastify()
+  fastify.register(fastifySwagger, { openapi: true, routePrefix: '/docs', exposeRoute: true })
+  fastify.addSchema({ ...schema, $id: 'requiredUniqueSchema' })
+  fastify.get('/', { schema: { query: { $ref: 'requiredUniqueSchema' } } }, () => {})
+  await fastify.ready()
+
+  const swaggerObject = fastify.swagger()
+  const api = await Swagger.validate(swaggerObject)
+  t.match(api.components.schemas['def-0'], schema)
 })
--- a/README.md
+++ b/README.md
@@ -211,13 +211,14 @@ An example of using `fastify-swagger` with `static` mode enabled can be found [h
  | hideUntagged       | false            | If `true` remove routes without tags from resulting Swagger/OpenAPI schema file.                                          |
  | initOAuth          | {}               | Configuration options for [Swagger UI initOAuth](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/).     |
  | openapi            | {}               | [OpenAPI configuration](https://swagger.io/specification/#oasObject).                                                     |
- | routePrefix        | '/documentation' | Overwrite the default Swagger UI route prefix.                                                                            |
+ | routePrefix         | '/documentation' | Overwrite the default Swagger UI route prefix.                                                                            |
  | staticCSP          | false            | Enable CSP header for static resources.                                                                                   |
  | stripBasePath      | true             | Strips base path from routes in docs.                                                                                     |
  | swagger            | {}               | [Swagger configuration](https://swagger.io/specification/v2/#swaggerObject).                                              |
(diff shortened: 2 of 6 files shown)


Eomm:
I think CI is failing due a temporary issue:

> throw new Error('Parallel webhook error:' + err + ', ' + JSON.stringify(data));

locally all is fine and 100% coverage reached

barticus:
It would be great to have some exported helper functions for use with `buildLocalReference`, e.g.

- BuildLocalReferencesSequentially (default)
- BuildLocalReferencesById (what you've shown you can do)

climba03003 (commented):
It means that we keep the original `$id` and move the resolution works to `title` field? Or the reverse way?

Eomm:
> It means that we keep the original $id and move the resolution works to title field? Or the reverse way?

No.

We add a `title` to fix the visual issue: `def-0` will never show up again reading the swagger documentation

The `$id` was and will not be changed

The `$ref` will continue to link relative schemas named `def-0`, `def-1` etc.. **but** it can be customized and removed by the user with the new configuration.

climba03003 (approved):
LGTM.

mcollina (approved):
lgtm

test/spec/openapi/schema.js
```diff
@@ -308,5 +308,25 @@ test('support global schema reference', async t => {
...
+  }
+  const fastify = Fastify()
+  fastify.register(fastifySwagger, { openapi: true, routePrefix: '/docs', exposeRoute: true })
+  fastify.addSchema({ ...schema, $id: 'requiredUniqueSchema' })
+  fastify.get('/', { schema: { query: { $ref: 'requiredUniqueSchema' } } }, () => {})
+  await fastify.ready()
+
+  const swaggerObject = fastify.swagger()
+  const api = await Swagger.validate(swaggerObject)
+  t.match(api.components.schemas['def-0'], schema)
```
peter-mouland, line 331: Why would this schema, which was added with an `$id`, not longer show the id via the `api`?

Eomm: It is explained in the first post of this pr

cesarquiroz:
Hey guys, I kind of have a similar issue, but Im using TypeScript and it looks like that `refResolver` wasn't exposed on the TypeScript definition, is there a way to get this fixed?

PR #472, opened 2021-09-20
OpenApi : nested ref in #/components/schemas/

### LINKED ISSUE #462: Allow nested $ref to resolve for OpenAPI?
### Prerequisites

### Issue

Currently, nested $refs are not resolved, $ref remains as the schema id.

If the id exists in components.schemas its should be rewritten to `#/components/schemas/{id of the schema}`

This can achieve this by adding modifying the loop at [/lib/spec/openapi/index.js#L31](https://github.com/fastify/fastify-swagger/blob/master/lib/spec/openapi/index.js#L31)

E.g search through each prop, and if a $ref is found with the same value as an existing schema, prefix it with  `#/components/schemas/`

```
Object.values(openapiObject.components.schemas)
.forEach((_) => {
    delete _.$id;
    if (_.properties) {
        Object.values(_.properties).forEach((prop) => {
            if (prop['$ref'] && typeof openapiObject.components.schemas[prop['$ref']] !== 'undefined') {
                prop['$ref'] = `#/components/schemas/${prop['$ref']}`;
            }
        });
    }
});
```
I would like to submit this as a pull request, but I'm not too familiar with swagger, JSON schema etc so would like some feedback before I do... Or should I just submit it a go from there?

briceruzand:
#### Checklist

- [X] tests are included

Fix #462

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -319,9 +319,22 @@ function prepareOpenapiMethod (schema, ref, openapiObject) {
   return openapiMethod
 }
 
+function prepareOpenapiSchemas (schemas, ref) {
+  return Object.entries(schemas)
+    .reduce((res, [name, schema]) => {
+      const _ = { ...schema }
+      const resolved = transformDefsToComponents(ref.resolve(_, { externalSchemas: [schemas] }))
+      return {
+        ...res,
+        [name]: resolved
+      }
+    }, {})
+}
+
 module.exports = {
   prepareDefaultOptions,
   prepareOpenapiObject,
   prepareOpenapiMethod,
+  prepareOpenapiSchemas,
   normalizeUrl
 }
--- a/lib/spec/openapi/index.js
+++ b/lib/spec/openapi/index.js
@@ -2,7 +2,7 @@
 
 const yaml = require('js-yaml')
 const { shouldRouteHide } = require('../../util/common')
-const { prepareDefaultOptions, prepareOpenapiObject, prepareOpenapiMethod, normalizeUrl } = require('./utils')
+const { prepareDefaultOptions, prepareOpenapiObject, prepareOpenapiMethod, prepareOpenapiSchemas, normalizeUrl } = require('./utils')
 
 module.exports = function (opts, cache, routes, Ref, done) {
   let ref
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

climba03003 (approved):
LGTM. 

But I would wait for @Eomm comment before merging. He has better understanding on how the `$ref` works on OpenAPI standard and how to resolve it correctly.

lib/spec/openapi/utils.js
```diff
@@ -319,9 +319,22 @@ function prepareOpenapiMethod (schema, ref, openapiObject) {
   return openapiMethod
 }
 
+function prepareOpenapiSchemas (schemas, ref) {
```
Eomm, line 322: Could you describe what you need to archive here?
I'm not sure to fully understand

With some testing this is not working if I put additional schemas in the bottom test:

```diff
    instance.addSchema({ $id: 'OrderItem', type: 'object', properties: { id: { type: 'integer' } } })
+    instance.addSchema({ $id: 'ProductItem', type: 'object', properties: { id: { type: 'integer' } } })
    instance.addSchema({ $id: 'Order', type: 'object', properties: { products: { type: 'array', items: { $ref: 'OrderItem' } } } })
    instance.post('/', { schema: { body: { $ref: 'Order' }, response: { 200: { $ref: 'Order' } } } }, () => {})
+    instance.post('/other', { schema: { body: { $ref: 'ProductItem' } } }, () => {})
```

briceruzand: Hi Ecomm, I d'like to replace local `$ref` with local  openapi components schema (prefixed by `#/components/schemas/`) in generated openapi components schema.
With that change, openapi components will be able to reference other local components.

I wrote a simple test of my use case (too simple may be) and I pick the function from [cjsewell@1a816ee](https://github.com/cjsewell/fastify-swagger/commit/1a816ee39c9e9a9d6e0a8e73d4705e7f68110c0f) which seems to work well on my test.

I going to improve my test (with a more complex case, has you done). In try too investigate why the test hang when you add other schema definitions.
Do you have a better idea on that function implementation?

Thx

zekth (commented):
Like @Eomm mentionned we need to add more test with complexe routing + schemas to try to detect any collision we might have.

lib/spec/openapi/utils.js
briceruzand: I have investigated on hanging test  you where face to.
I seems to be cause by unfinished fastify-swagger initialization whereas fastify `fastify.ready()` is ok.
I have rewrote test using async/await to provide more readable tests and I never have this trouble anymore.

Hanging test behavior also exists on your test case without pull request feature.

briceruzand:
@zekth I provide a more complex test case of nested ref in schema definitions.

https://github.com/fastify/fastify-swagger/blob/2f28ad6320d8da35bc6236942ebfda63d011acf7/test/spec/openapi/refs.js#L60-L86
Thx for your review.

Best regards

zekth (approved):
Good job on this.

Eomm (approved):

briceruzand:
Thank you all

PR #478, opened 2021-09-26
fix: nested refs

peter-mouland:
Start work towards https://github.com/fastify/fastify-swagger/issues/462

TBC:
 - [ ] how to handle ids/refs that arent modified using `buildLocalReference`?
     - Current behaviour `t.equal(definitions['def-3'].properties.b.$ref, '#/definitions/schemaB')`, vs
     - more correct `t.equal(definitions['def-3'].properties.b.$ref, '#/definitions/def-2')`?
 - [ ] de we need to code for `key === 'examples'` in the swagger spec?

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #479, opened 2021-09-27
fix: nested swagger ref

peter-mouland:
related to https://github.com/fastify/fastify-swagger/pull/478 and https://github.com/fastify/fastify-swagger/pull/472

Fixes nested + complex ref's when using the swagger spec

#### Checklist

### DIFF
--- a/lib/spec/swagger/index.js
+++ b/lib/spec/swagger/index.js
@@ -2,7 +2,7 @@
 
 const yaml = require('js-yaml')
 const { shouldRouteHide } = require('../../util/common')
-const { prepareDefaultOptions, prepareSwaggerObject, prepareSwaggerMethod, normalizeUrl } = require('./utils')
+const { prepareDefaultOptions, prepareSwaggerObject, prepareSwaggerMethod, normalizeUrl, prepareSwaggerDefinitions } = require('./utils')
 
 module.exports = function (opts, cache, routes, Ref, done) {
   let ref
@@ -19,16 +19,10 @@ module.exports = function (opts, cache, routes, Ref, done) {
     const swaggerObject = prepareSwaggerObject(defOpts, done)
 
     ref = Ref()
-    swaggerObject.definitions = {
+    swaggerObject.definitions = prepareSwaggerDefinitions({
       ...swaggerObject.definitions,
       ...(ref.definitions().definitions)
-    }
-
-    // Swagger doesn't accept $id on /definitions schemas.
-    // The $ids are needed by Ref() to check the URI so we need
-    // to remove them at the end of the process
-    Object.values(swaggerObject.definitions)
-      .forEach(_ => { delete _.$id })
+    }, ref)
 
     swaggerObject.paths = {}
     for (const route of routes) {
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

peter-mouland:
hey @climba03003  @Eomm 
I've rebased from master to remove the conflicts, is there anything else we need to do to get the PR in?

lib/spec/swagger/index.js
```diff
@@ -19,16 +19,19 @@ module.exports = function (opts, cache, routes, Ref, done) {
...
       ...(ref.definitions().definitions)
-    }
+    }, ref)
 
     // Swagger doesn't accept $id on /definitions schemas.
     // The $ids are needed by Ref() to check the URI so we need
     // to remove them at the end of the process
     Object.values(swaggerObject.definitions)
-      .forEach(_ => { delete _.$id })
+      .forEach(_ => {
```
climba03003, line 31: Can you move this logic inside `prepareSwaggerDefinitions`? I think it is not necessary to have two loop here.

climba03003 (changes requested):

lib/spec/swagger/index.js
peter-mouland: done - and i moved the openAPI version to the same equivalent place

climba03003: Thanks

lib/spec/swagger/utils.js
```diff
@@ -266,9 +266,29 @@ function prepareSwaggerMethod (schema, ref, swaggerObject) {
...
+      const _ = { ...definition }
+      const resolved = ref.resolve(_, { externalSchemas: [definitions] })
+
+      // Swagger doesn't accept $id on /definitions schemas.
+      // The $ids are needed by Ref() to check the URI so we need
+      // to remove them at the end of the process
+      delete resolved.$id
+      delete resolved.definitions
+
+      return {
```
climba03003, line 281: I do not catch it in first place. But we should prevent a `reduce` loop and `spread` operator in here.
It create a new object each round of the loop.

```js
res[name] = resolved
return res
```

The above code should be enough.

climba03003: Benchmark

```js
const { performance } = require('perf_hooks');
const arr = Array(10000).fill(0)

function speedSpread() {
  const start = performance.now()
  const obj = arr.reduce(function(obj) {
    return { ...obj, [Math.random()]: Math.random() } 
  }, {})
  return performance.now() - start
}

function speedPropForEach() {
  const start = performance.now()
  const obj = {}
  arr.forEach(function() {
    obj[Math.random()] = Math.random()
  })
  return performance.now() - start
}

function speedPropReduce() {
  const start = performance.now()
  const obj = arr.reduce(function(obj) {
    obj[Math.random()] = Math.random()
    return obj
  }, {})
  return performance.now() - start
}

const a = speedSpread()
const b = speedPropForEach()
const c = speedPropReduce()

console.log('speedSpread', a)
console.log('speedPropForEach', b)
console.log('speedPropReduce', c)
```

climba03003 (changes requested):

lib/spec/swagger/utils.js
peter-mouland: nice - updated for openAPI and swagger 👍

climba03003 (approved):
LGTM.

PR #480, opened 2021-09-27
fix: generate correct openapi with response schema containing ref

davidebianchi:
When a response schema is created with a `$ref` not at first level (as in the added test), the openapi spec is not valid because of the `$id` key from the added schema.

Here the output of the test before (not a valid oas)

```yml
openapi: 3.0.3
info:
  version: 4.12.1
  title: fastify-swagger
components:
  schemas:
    order:
      type: string
      enum:
        - foo
paths:
  /:
    post:
      responses:
        '200':
          description: Default Response
          content:
            application/json:
              schema:
                type: object
                properties:
                  order:
                    $id: order
                    type: string
                    enum:
                      - foo
```

and after the change (this is valid because order does not contains the `$id` key):

```yml
openapi: 3.0.3
info:
  version: 4.12.1
  title: fastify-swagger
components:
  schemas:
    order:
      type: string
      enum:
        - foo
paths:
  /:
    post:
      responses:
        '200':
          description: Default Response
          content:
            application/json:
              schema:
                type: object
                properties:
                  order:
                    type: string
                    enum:
                      - foo
```

#### Checklist

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -88,6 +88,8 @@ function transformDefsToComponents (jsonSchema) {
       } else if (key === 'examples' && Array.isArray(jsonSchema[key]) && (jsonSchema[key].length === 1)) {
         jsonSchema.example = jsonSchema[key][0]
         delete jsonSchema[key]
+      } else if (key === '$id') {
+        delete jsonSchema[key]
       } else {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Eomm (approved):

PR #520, opened 2021-12-27
OpenApi: Fix params mapping on relative URL

fatal10110:
Fix for a bug, when `stripBasePath` is disabled and openApi is configured. it will not map path params to fastify params.

#### Checklist

### DIFF
--- a/examples/options.js
+++ b/examples/options.js
@@ -63,6 +63,41 @@ const openapiOption = {
   }
 }
 
+const openapiRelaiveOptions = {
+  openapi: {
+    info: {
+      title: 'Test swagger',
+      description: 'testing the fastify swagger api',
+      version: '0.1.0'
+    },
+    servers: [
+      {
+        url: '/test'
+      }
+    ],
+    tags: [
+      { name: 'tag' }
+    ],
+    components: {
+      securitySchemes: {
+        apiKey: {
+          type: 'apiKey',
+          name: 'apiKey',
+          in: 'header'
+        }
+      }
+    },
+    security: [{
+      apiKey: []
+    }],
+    externalDocs: {
+      description: 'Find more info here',
+      url: 'https://swagger.io'
+    }
+  },
+  stripBasePath: false
+}
+
 const schemaQuerystring = {
   schema: {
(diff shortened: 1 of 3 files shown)


test/spec/openapi/route.js
```diff
@@ -7,6 +7,7 @@ const yaml = require('js-yaml')
 const fastifySwagger = require('../../../index')
 const {
   openapiOption,
+  openapiRelaiveOptions,
```
mcollina, line 10: why this change?

examples/options.js
```diff
@@ -250,6 +285,7 @@ const schemaOperationId = {
 
 module.exports = {
   openapiOption,
+  openapiRelaiveOptions,
```
mcollina, line 288: what does relaive mean? Is it a typo?

mcollina (commented):
I think the code should be linted.

examples/options.js
fatal10110: Thanks, will fix it

test/spec/openapi/route.js
fatal10110: Im using it in a test to, these are the options im able to reproduce the bug, so I thought this is the best way to test it

fatal10110:
> I think the code should be linted.

Not sure what do you mean by "linted", I used `npm run test` which is also triggering the "standard" cli, also the CI checks passed

lib/spec/openapi/utils.js
```diff
@@ -67,7 +67,7 @@ function prepareOpenapiObject (opts) {
 }
 
 function normalizeUrl (url, servers, stripBasePath) {
-  if (!stripBasePath) return url
+  if (!stripBasePath) return formatParamUrl(url)
```
climba03003, line 70: Good catch

mcollina:
> Not sure what do you mean by "linted", I used npm run test which is also triggering the "standard" cli, also the CI checks passed

nevermind, things were scrambled on mobile.

Eomm (approved):

climba03003 (approved):

PR #522, opened 2022-01-07
feat(openapi): support const (#521)

### LINKED ISSUE #521: using `const` keyword produce invalid swagger file
### Prerequisites

### Fastify version

3.25.0

### Plugin version

4.13.0

### Node.js version

16.13.0

### Operating system

macOS

10.15.7 (19H2)

### Description

The swagger plugin produces an invalid swagger file when using the json-schema `const` keyword ([spec](http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.1.3))

### Steps to Reproduce

```
const requestBodySchema = {
  type: 'object',
  required: ['prop'],
  properties: {
    prop: { const: 'my-const' },
  },
}

export default async (fastify) => {
  fastifty.route({
  method: 'POST',
  url: '/',
  schema: {
    body: requestBodySchema,
  },
  handler,
});
};
```
produce:
```
...
schema:
  type: object
  required:
    - prop
  properties:
    prop:
      const: my-const
...
```
Swagger validator error message:
```
Structural error at paths./v1/users/.post.requestBody.content.application/json.schema.properties.prop
should NOT have additional properties
additionalProperty: const
```

### Expected Behavior

The swagger plugin will generate a valid swagger file, for example, enum with a single value when using json-schema `const` keyword

mikicho:
close #521 

#### Checklist

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -84,6 +84,10 @@ function transformDefsToComponents (jsonSchema) {
     if (jsonSchema.patternProperties) {
       jsonSchema.additionalProperties = { type: 'string' }
       delete jsonSchema.patternProperties
+    } else if (jsonSchema.const) {
+      // Handle const, that is not part of OpenAPI definitions
+      jsonSchema.enum = [jsonSchema.const]
+      delete jsonSchema.const
     }
 
     Object.keys(jsonSchema).forEach(function (key) {
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -85,7 +85,8 @@ function transformDefsToComponents (jsonSchema) {
       jsonSchema.additionalProperties = { type: 'string' }
       delete jsonSchema.patternProperties
     } else if (jsonSchema.const) {
-      // Handle const, that is not part of OpenAPI definitions
+      // OAS 3.1 supports `const` but it is not supported by `swagger-ui`
+      // https://swagger.io/docs/specification/data-models/keywords/
       jsonSchema.enum = [jsonSchema.const]
       delete jsonSchema.const
     }
--- a/lib/spec/swagger/utils.js
+++ b/lib/spec/swagger/utils.js
@@ -165,7 +165,8 @@ function replaceUnsupported (jsonSchema) {
       jsonSchema.additionalProperties = { type: 'string' }
       delete jsonSchema.patternProperties
     } else if (jsonSchema.const) {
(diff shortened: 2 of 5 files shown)


lib/spec/openapi/utils.js
```diff
@@ -84,6 +84,10 @@ function transformDefsToComponents (jsonSchema) {
     if (jsonSchema.patternProperties) {
       jsonSchema.additionalProperties = { type: 'string' }
       delete jsonSchema.patternProperties
+    } else if (jsonSchema.const) {
+      // Handle const, that is not part of OpenAPI definitions
```
Eomm, line 88:
```suggestion
      // OAS 3.1 supports `const` but it is not supported by `swagger-ui`
      // https://swagger.io/docs/specification/data-models/keywords/
```

Eomm (changes requested):
Could you add a unit test for this feature?

mikicho:
@Eomm 
> Could you add a unit test for this feature?

Not sure we need to, `swagger/openapi should return valid swagger object` already testing it and it seems like the general approach is not test the transformer itself (the output..)

mcollina (commented):

lib/spec/swagger/utils.js
```diff
@@ -164,6 +164,11 @@ function replaceUnsupported (jsonSchema) {
     if (jsonSchema.patternProperties) {
       jsonSchema.additionalProperties = { type: 'string' }
       delete jsonSchema.patternProperties
+    } else if (jsonSchema.const) {
+      // OAS 3.1 supports `const` but it is not supported by `swagger-ui`
+      // https://swagger.io/docs/specification/data-models/keywords/
```
Eomm, line 169: The old comment, here in the `swagger` directory was fine.

- Swagger is an alias for OAS version 2 that does not support `const`
- The other comment, in `openapi` folder, is the file to support OAS version 3. So the comment there must be different from the following

```suggestion
      // Handle const, that is not part of OpenAPI definitions
```

Eomm (changes requested):
We need a test that uses the `const` keyword within a schema and check that the output produced is the `enum` instead.

This use case is not covered by the existing tests

lib/spec/swagger/utils.js
mikicho: thanks, fixed

Eomm (approved):

mikicho:
@Eomm can you please nerge it?

mcollina (approved):
lgtm

darkgl0w (approved):
LGTM

Issue #555, opened 2022-03-01
enhance url transformation and hiding options


samzilverberg:
Hi there, I added a feature proposal according to your template.
If you think it'd be useful (with or without some changes), let me know and I can try and work on a PR for this.

### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

currently, the plugin supports hiding routes (line 38) and removing base path if it's part of the server path, before adding the route to the `paths` area of the spec.

https://github.com/fastify/fastify-swagger/blob/146bfb112322492b63b0927bd953ad8cde704241/lib/spec/openapi/index.js#L38-L42

I'd really like to be able to transform the url a little more, and another nice to have would be being able to hide routes based on same custom condition (predicate on the url).

### Motivation

the usecase that I'd like to support is when fastify is behing a reverse proxy and the defined routes are not the final urls that should appear in the api spec (for users that use the api through the reverse proxy).

example fastify defined routes and the wanted transformation:

- `get /public/blabla` : should be `get /api/blabla`
- `get /private/blabla` : should not be in spec
- `get /health`: should not be in spec

It'd be nice to be able to
- transform the prefix `/public` to `/api` (or more general trasnformation fn).
- hide urls with prefix `/private`, or hide anything that's not '/public', and not need to worry about tagging or using the hide flag.

### Example

example usage could be:

```
fastify.register(require('fastify-swagger'), {
  swagger: { ... },
  ...
  hideRouteFn: url => {
    return url.startsWith('/private')
  },  
  transformUrl: url => {
   return url.replace(/^(/public)/,"/api");
  }
```

zekth:
That's a good use case an approach. I'm +1 on this. Could you write a PR for this?

The use case you're describing with the reverse proxy is an interesting approach. Is this correct?
```
get /public/blabla : should be get /api/blabla (in prod)
get /private/blabla : should not be in spec (in prod)
get /health: should not be in spec (in prod)
```

samzilverberg:
yes i'll try to find time in the next couple of days to do this.
From the code I've seen so far I don't think it'd be too complicated to add this.
we can talk about option naming and other fixes in the PR 

The use case I'm having is not env specific.
our apps are behind a k8s ingress controller (an nginx reverse proxy under the hood), and only routes that start with a specific prefix are "public" to the world. 
the rest are internal routes.

[...]

PR #559, opened 2022-03-06
Option to hide route using predicate

samzilverberg:
adds the hide predicate option discussed in https://github.com/fastify/fastify-swagger/issues/555

I added typescript typings but didn't find any useful documentation on how to actually test them.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,15 +219,16 @@ An example of using `fastify-swagger` with `static` mode enabled can be found [h
  | exposeRoute        | false            | Exposes documentation route.                                                                                              |
  | hiddenTag          | X-HIDDEN         | Tag to control hiding of routes.                                                                                          |
  | hideUntagged       | false            | If `true` remove routes without tags from resulting Swagger/OpenAPI schema file.                                          |
+ | hideRoutePredicate | null             | predicate function (url, schema) => boolean to decide if should hide route                                                |
  | initOAuth          | {}               | Configuration options for [Swagger UI initOAuth](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/).     |
  | openapi            | {}               | [OpenAPI configuration](https://swagger.io/specification/#oasObject).                                                     |
- | routePrefix         | '/documentation' | Overwrite the default Swagger UI route prefix.                                                                            |
+ | routePrefix        | '/documentation' | Overwrite the default Swagger UI route prefix.                                                                            |
  | staticCSP          | false            | Enable CSP header for static resources.                                                                                   |
  | stripBasePath      | true             | Strips base path from routes in docs.                                                                                     |
  | swagger            | {}               | [Swagger configuration](https://swagger.io/specification/v2/#swaggerObject).                                              |
  | transform          | null             | Transform method for schema.                                                                                              |
- | transformStaticCSP | undefined         | Synchronous function to transform CSP header for static resources if the header has been previously set.                  |
- | uiConfig            | {}               | Configuration options for [Swagger UI](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md). Must be literal values, see [#5710](https://github.com/swagger-api/swagger-ui/issues/5710).|
+ | transformStaticCSP | undefined        | Synchronous function to transform CSP header for static resources if the header has been previously set.                  |
+ | uiConfig           | {}               | Configuration options for [Swagger UI](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md). Must be literal values, see [#5710](https://github.com/swagger-api/swagger-ui/issues/5710).|
  | uiHooks            | {}               | Additional hooks for the documentation's routes. You can provide the `onRequest` and `preHandler` hooks with the same [route's options](https://www.fastify.io/docs/latest/Routes/#options) interface.|
  | refResolver        | {}               | Option to manage the `$ref`s of your application's schemas. Read the [`$ref` documentation](#register.options.refResolver) |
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -219,15 +219,16 @@ An example of using `fastify-swagger` with `static` mode enabled can be found [h
...
  | swagger            | {}               | [Swagger configuration](https://swagger.io/specification/v2/#swaggerObject).                                              |
  | transform          | null             | Transform method for schema.                                                                                              |
- | transformStaticCSP | undefined         | Synchronous function to transform CSP header for static resources if the header has been previously set.                  |
- | uiConfig            | {}               | Configuration options for [Swagger UI](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md). Must be literal values, see [#5710](https://github.com/swagger-api/swagger-ui/issues/5710).|
+ | transformStaticCSP | undefined        | Synchronous function to transform CSP header for static resources if the header has been previously set.                  |
+ | uiConfig           | {}               | Configuration options for [Swagger UI](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md). Must be literal values, see [#5710](https://github.com/swagger-api/swagger-ui/issues/5710).|
```
samzilverberg, line 231: just did some allignments of the '|'

README.md
```diff
@@ -219,15 +219,16 @@ An example of using `fastify-swagger` with `static` mode enabled can be found [h
...
  | hiddenTag          | X-HIDDEN         | Tag to control hiding of routes.                                                                                          |
  | hideUntagged       | false            | If `true` remove routes without tags from resulting Swagger/OpenAPI schema file.                                          |
+ | hideRoutePredicate | null             | predicate function (url, schema) => boolean to decide if should hide route                                                |
  | initOAuth          | {}               | Configuration options for [Swagger UI initOAuth](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/).     |
  | openapi            | {}               | [OpenAPI configuration](https://swagger.io/specification/#oasObject).                                                     |
- | routePrefix         | '/documentation' | Overwrite the default Swagger UI route prefix.                                                                            |
+ | routePrefix        | '/documentation' | Overwrite the default Swagger UI route prefix.                                                                            |
```
samzilverberg, line 225: just did some allignments of the '|'

README.md
```diff
@@ -647,9 +648,17 @@ fastify.get('/user/:id/address', {
...
 - Pass `{ hide: true }` to the schema object inside the route declaration.
 - Use the tag declared in `hiddenTag` options property inside the route declaration. Default is `X-HIDDEN`.
+- use `hideRoutePredicate` option to pass a predicate fn (url, schema) => boolean to decide if a route should be hidden.
+  - note: if the `transform` option is used then the transformed schema is the one passed to this predicate 
+
+hiding options cascade: schema hide -> tag -> predicate, returning on the first "true" found.
+
+example: if a route is hidden through the schema hide flag then the other hide options will be ignored.
+
+another example: if the route is hidden through a tag then the hide predicate will be ignored.
```
samzilverberg, line 661: let me know if you have a better way to describe this.

README.md
```diff
@@ -219,15 +219,16 @@ An example of using `fastify-swagger` with `static` mode enabled can be found [h
  | exposeRoute        | false            | Exposes documentation route.                                                                                              |
  | hiddenTag          | X-HIDDEN         | Tag to control hiding of routes.                                                                                          |
  | hideUntagged       | false            | If `true` remove routes without tags from resulting Swagger/OpenAPI schema file.                                          |
+ | hideRoutePredicate | null             | predicate function (url, schema) => boolean to decide if should hide route                                                |
```
samzilverberg, line 222: name suggestions are welcomed :)

mcollina: I'm not really convinced by the name, why "predicate"?

samzilverberg: Because it's common enough to use `predicate` when refering to a function that returns true or false.
here's a result from google with more details:

https://www.baeldung.com/cs/predicates

you're more than welcome to suggest any other name, i'm open to changes!

I also considered just calling it `hideRouteFn` but from my own experience shortening function to 'Fn' is not very welcome on libraries.

climba03003 (changes requested):
Instead of adding a new function, I would like it to improve the existing `transform` function. It should allow to pass in the `schema` and `route` then mutate all of them inside the single function.

samzilverberg:
@climba03003 I also think it'd nicer if `transform` does both.
I had this though initially but was afraid it would introduce a big breaking change.
let me take a deeper look into it and see if it's not as bad as i think.

samzilverberg:
@climba03003 also: you meant to comment with this about url transformation, right? 
and by mistake put it here on the `hide` related PR? or not?

climba03003:
> @climba03003 also: you meant to comment with this about url transformation, right? and by mistake put it here on the `hide` related PR? or not?

I think it should do both of them within a single transform. Since, transform can change the schema. It only need to allow change of url for another feature.

samzilverberg:
@climba03003 I think I understand what you're suggestion.
transform could perform whatever checks it wants with url & schema and return a transformed url & schema accordingly.

to hide: it can add `hide` flag on schema or the hide tag

I like this idea, will try to impl it.

climba03003:
> @climba03003 I think I understand what your suggestion. transform could perform whatever checks it wants with url & schema and return a transformed url & schema accordingly.
> 
> to hide: it can add `hide` flag on schema or the hide tag
> 
> I like this idea, will try to impl it.
> 
> plz correct me if i got it wrong

Yes, you got my idea.

samzilverberg:
closing in favor of https://github.com/fastify/fastify-swagger/pull/562

Issue #501, opened 2021-11-09
Add missing property refResolver to types file (index.d.ts)


pharindoko:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I use typebox and fastify swagger in combination

One thing that I`m not happy with is the def-# stuff in the schema name.
I use fastify.addSchema function to add the type...

![image](https://user-images.githubusercontent.com/5619511/141019081-9827d7fa-f1e7-49b1-8ed1-0cc43dfab739.png)

<img width="616" alt="image" src="https://user-images.githubusercontent.com/5619511/141018547-1c64f262-f4ae-415d-a5f6-6097cebe31ae.png">

What I would like to have:
Be able to set a specific name for the schema e.g. in this case components/schema/UserObject

### Motivation

I would like to generate my types for the frontend from the openapi spec.
But Def-0 as type is not helpful...
![image](https://user-images.githubusercontent.com/5619511/141018973-b311ec82-d962-4626-9b18-fc971a77e7b4.png)

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

pharindoko:
Hey @mcollina,

hmm haven`t seen this.
Guess that fixes my problem...
https://github.com/fastify/fastify-swagger/blob/1a68013d9a3f02034b48b87a4902040f113a0cfa/README.md#managing-your-refs

Need to try...

pharindoko:
seems like the refResolver is missing in the typescript index.d.ts file at this place:
https://github.com/fastify/fastify-swagger/blob/dc8e29acc488e22f6fdebcfdc64f5d61e3863a63/index.d.ts#L118

pharindoko:
ok adding this to index.d.ts file in line 119 immediately fixed the problem but it would be nice use a better type than Object ...
```typescript
  refResolver?: Object; 
```

pharindoko:
Created a repo where I fixed this issue with an interface extension...
https://github.com/pharindoko/fastify-prisma-ts/blob/8f43f86b12019ce8f0754cf491f121339d564e5b/index.ts#L14

@mcollina can you add the missing typing for the refResolver dependency ?

Eomm:
This is the `refResolver` API: https://github.com/Eomm/json-schema-resolver#usage-resolve-one-schema-against-external-schemas

Would you mind adding it?

[...]

PR #561, opened 2022-03-06
refResolver index.d.ts typescript typings

samzilverberg:
resolves https://github.com/fastify/fastify-swagger/issues/501

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -103,6 +103,18 @@ export type FastifySwaggerInitOAuthOptions = Partial<{
   usePkceWithAuthorizationCodeGrant: boolean
 }>
 
+type JSONValue =
+    | string
+    | null
+    | number
+    | boolean
+    | JSONObject
+    | Array<JSONValue>;
+
+interface JSONObject {
+  [key: string]: JSONValue;
+}
+
 export interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
   mode?: 'dynamic';
   swagger?: Partial<OpenAPIV2.Document>;
@@ -118,6 +130,31 @@ export interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
    * Overwrite the route schema
    */
   transform?: Function;
+
+  refResolver?: {
+    /** Clone the input schema without changing it. Default to `false`. */
+    clone?: boolean;
+    buildLocalReference: (
+      /** The `json` that is being resolved. */
+      json: JSONObject,
+      /** The `baseUri` object of the schema. */
+      baseUri: {
+        scheme?: string;
+        userinfo?: string;
+        host?: string;
+        port?: number | string;
+        path?: string;
+        query?: string;
+        fragment?: string;
+        reference?: string;
+        error?: string;
+      },
+      /** `fragment` is the `$ref` string when the `$ref` is a relative reference. */
+      fragment: string,
+      /** `i` is a local counter to generate a unique key. */
+      i: number
+    ) => string;
+  }
 }
 
 export interface StaticPathSpec {
--- a/test/types/types.test.ts
+++ b/test/types/types.test.ts
@@ -51,7 +51,10 @@ const fastifyDynamicSwaggerOptions: SwaggerOptions = {
   exposeRoute: true,
   hiddenTag: 'X-HIDDEN',
   hideUntagged: true,
-  stripBasePath: true
+  stripBasePath: true,
+  refResolver: {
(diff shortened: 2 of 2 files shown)


darkgl0w:
Can you add a unit test for this typescript definition please ?

samzilverberg:
added test.

index.d.ts
```diff
@@ -118,6 +118,10 @@ export interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
    * Overwrite the route schema
    */
   transform?: Function;
+
+  refResolver?: {
+    buildLocalReference: (json: any, baseUri: any, fragment: string, i: number) => string
```
darkgl0w, line 123:
```suggestion
    /** Clone the input schema without changing it. Default to `false`. */
    clone?: boolean;
    buildLocalReference: (
      /** The `json` that is being resolved. */
      json: any,
      /** The `baseUri` object of the schema. */
      baseUri: {
        scheme?: string;
        userinfo?: string;
        host?: string;
        port?: number | string;
        path?: string;
        query?: string;
        fragment?: string;
        reference?: string;
        error?: string;
      },
      /** `fragment` is the `$ref` string when the `$ref` is a relative reference. */
      fragment: string,
      /** `i` is a local counter to generate a unique key. */
      i: number
    ) => string;
```

It was missing the clone option, and baseUri is the parse result from `uri-js`.

darkgl0w (commented):
Nitpicks.

the `json` property can be typed as `Record<string, any>` I think. Can you explore this way ?

EDIT: Maybe we can type JSON Objects like this :

```ts
type JSONValue = boolean | Array<JSONValue> | number | null | string | Record<string, JSONValue>;

interface JSONObject extends Record<string, JSONValue> {}
```

mcollina (approved):
lgtm

samzilverberg:
@darkgl0w re using a `JSONObject` :

lib users that need to use `refResolver` would have to go the the documentation and figure out what `json` is and what fields it contains (it's a pretty flexible schema json AFAIK).
The compiler and the added `JSONObject` types are not going to be much help with that.
I think `any` is a good first step and that a more accurate typing can be added later from someone that has better knowledge of `refResolver` (or has more time to explore).

darkgl0w:
@samzilverberg I gave you the typescript definition for `JSONObject`. You just need to add this :
```ts
export type JSONValueType =
  | boolean
  | JSONObject
  | JSONValueType[]
  | null
  | number
  | string;
export interface JSONObject {
  [key: string]: JSONValueType;
}
```

and then change the `json` type from `any` to `JSONObject` as it covers all possible values. :smiley: 
With this users can type their `json` objects by importing the `JSONObject` interface and if needed they can also import the `JSONValue` type. ^^

samzilverberg:
i added a json type for the `json` field

index.d.ts
```diff
@@ -118,6 +125,31 @@ export interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
    * Overwrite the route schema
    */
   transform?: Function;
+
+  refResolver?: {
+    /** Clone the input schema without changing it. Default to `false`. */
+    clone?: boolean;
+    buildLocalReference: (
+      /** The `json` that is being resolved. */
+      json: JSONValue,
```
darkgl0w, line 134: `'hello'`, `1`, `true` ... etc are valid according to this type.

darkgl0w (changes requested):

index.d.ts
samzilverberg: ahh i see. i'll add the JSONObject type as well :)

index.d.ts
```diff
@@ -103,6 +103,17 @@ export type FastifySwaggerInitOAuthOptions = Partial<{
   usePkceWithAuthorizationCodeGrant: boolean
 }>
 
+type JSONValue =
+    | string
+    | number
```
darkgl0w, line 108:
```suggestion
    | null
    | number
```
`null` is a valid type too for a json object ^^

darkgl0w (approved):
LGTM with the suggested change. ^^

index.d.ts
samzilverberg: ty for spotting that. the trouble with making changes late in the evening is you (me) miss the small details

samzilverberg:
@darkgl0w  / @mcollina i'm not sure if I can merge or if you have to do it it.
other than that, github shows me a warning saying that a maintainer (one of you?) needs to approve a workflow.

climba03003 (approved):

PR #560, opened 2022-03-06
transformUrl option

samzilverberg:
adds the url transformation option discussed in https://github.com/fastify/fastify-swagger/issues/555

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #562, opened 2022-03-06
feat(major): transform add url to option and document more possible use cases

samzilverberg:
resolves issue https://github.com/fastify/fastify-swagger/issues/555

combines the following PRs into one by enhancing the existing `transform` option:

https://github.com/fastify/fastify-swagger/pull/559
https://github.com/fastify/fastify-swagger/pull/560

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,19 +262,19 @@ fastify.register(require('fastify-swagger'), {
   swagger: { ... },
   transform: ({ schema, url }) => {
     const {
-      params = undefined,
-      body = undefined,
-      querystring = undefined,
-      headers = undefined,
-      response = undefined,
-      ...rest
+      params,
+      body,
+      querystring,
+      headers,
+      response,
+      ...restOfFastifySchema
     } = schema
-    const transformedSchema = { ...rest }
+    const transformedSchema = { ...restOfFastifySchema }
     let transformedUrl = url
 
     // Transform the schema as you wish with your own custom logic.
     // In this example convert is from 'joi-to-json' lib and converts a Joi based schema to json schema
-    if (params) transfotransformedSchemarmed.params = convert(params)
+    if (params) transformedSchema.params = convert(params)
     if (body) transformedSchema.body = convert(body)
     if (querystring) transformedSchema.querystring = convert(querystring)
     if (headers) transformedSchema.headers = convert(headers)
--- a/README.md
+++ b/README.md
@@ -267,9 +267,8 @@ fastify.register(require('fastify-swagger'), {
       querystring,
       headers,
       response,
-      ...restOfFastifySchema
     } = schema
-    const transformedSchema = { ...restOfFastifySchema }
+    const transformedSchema = Object.assign({}, schema) //shallow copy of schema
     let transformedUrl = url
 
     // Transform the schema as you wish with your own custom logic.
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ An example of using `fastify-swagger` with `static` mode enabled can be found [h
  | staticCSP          | false            | Enable CSP header for static resources.                                                                                   |
  | stripBasePath      | true             | Strips base path from routes in docs.                                                                                     |
  | swagger            | {}               | [Swagger configuration](https://swagger.io/specification/v2/#swaggerObject).                                              |
- | transform          | null             | Transform method for schema.                                                                                              |
+ | transform          | null             | Transform method for the route's schema and url. [documentation](#register.options.transform).                                                                                              |
  | transformStaticCSP | undefined         | Synchronous function to transform CSP header for static resources if the header has been previously set.                  |
  | uiConfig            | {}               | Configuration options for [Swagger UI](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md). Must be literal values, see [#5710](https://github.com/swagger-api/swagger-ui/issues/5710).|
  | uiHooks            | {}               | Additional hooks for the documentation's routes. You can provide the `onRequest` and `preHandler` hooks with the same [route's options](https://www.fastify.io/docs/latest/Routes/#options) interface.|
@@ -241,30 +241,54 @@ If you set `exposeRoute` to `true` the plugin will expose the documentation with
 | `'/documentation/*'`    | External files that you may use in `$ref`  |
 
 <a name="register.options.transform"></a>
-#### Transforms
+#### Transform
 
-To use different schemas such as [Joi](https://github.com/hapijs/joi) you can pass a synchronous `transform` method in the options to convert them back to standard JSON schemas expected by this plugin to generate the documentation (`dynamic` mode only).
+By passing a synchronous `transform` function you can modify the route's url and schema.
+
+Some possible uses of this are:
+
+- add the `hide` flag on schema according to your own logic based on url & schema
+- altering the route url into something that's more suitable for the api spec
+- using different schemas such as [Joi](https://github.com/hapijs/joi) and transforming them to standard JSON schemas expected by this plugin
+
+This option is available in `dynamic` mode only.
+
+Examples of all the possible uses mentioned:
 
 ```js
 const convert = require('joi-to-json')
 
 fastify.register(require('fastify-swagger'), {
   swagger: { ... },
-  ...
-  transform: schema => {
+  transform: ({ schema, url }) => {
     const {
       params = undefined,
       body = undefined,
       querystring = undefined,
-      ...others
+      headers = undefined,
+      response = undefined,
+      ...rest
     } = schema
-    const transformed = { ...others }
-    if (params) transformed.params = convert(params)
-    if (body) transformed.body = convert(body)
-    if (querystring) transformed.querystring = convert(querystring)
-    return transformed
+    const transformedSchema = { ...rest }
+    let transformedUrl = url
+
+    // Transform the schema as you wish with your own custom logic.
+    // In this example convert is from 'joi-to-json' lib and converts a Joi based schema to json schema
+    if (params) transfotransformedSchemarmed.params = convert(params)
+    if (body) transformedSchema.body = convert(body)
+    if (querystring) transformedSchema.querystring = convert(querystring)
+    if (headers) transformedSchema.headers = convert(headers)
+    if (response) transformedSchema.response = convert(response)
+
+    // can add the hide tag if needed
+    if (url.startsWith('/internal')) transformedSchema.hide = true
+
+    // can transform the url
+    if (url.startsWith('/latest_version/endpoint')) transformedUrl = url.replace('latest_version', 'v3')
+
+    return { schema: transformedSchema, url: transformedUrl }
   }
-}
+})
 ```
 
 <a name="register.options.refResolver"></a>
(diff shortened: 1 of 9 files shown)


mcollina (approved):
lgtm

README.md
```diff
@@ -225,7 +225,7 @@ An example of using `fastify-swagger` with `static` mode enabled can be found [h
  | staticCSP          | false            | Enable CSP header for static resources.                                                                                   |
  | stripBasePath      | true             | Strips base path from routes in docs.                                                                                     |
  | swagger            | {}               | [Swagger configuration](https://swagger.io/specification/v2/#swaggerObject).                                              |
- | transform          | null             | Transform method for schema.                                                                                              |
+ | transform          | null             | Transform method for the route's schema and url: `(schema: any, url: string) => { schema: any, url: string }`.                                                                                              |
```
samzilverberg, line 228: **IMPORTANT: this is a breaking change that requires a major version release**

it changes the function return value.
so any existing implementations will have to be migrated from returning simple `return transformed` to `return { schema: transformed, url }`

samzilverberg: we can add some fallback to support the old code (and release as minor), but it'l be a little ugly.
fallback code is usually ugly  😆

lib/spec/openapi/index.js
```diff
@@ -26,18 +26,20 @@ module.exports = function (opts, cache, routes, Ref, done) {
...
+      const schema = transformResult.schema ?? route.schema
       const shouldRouteHideOpts = {
         hiddenTag: defOpts.hiddenTag,
         hideUntagged: defOpts.hideUntagged
       }
 
       if (shouldRouteHide(schema, shouldRouteHideOpts)) continue
 
-      const url = normalizeUrl(route.url, defOpts.servers, defOpts.stripBasePath)
+      let url = transformResult.url ?? route.url
```
climba03003, line 41: Nullish operator add on `node@14` and we support `node@10` currently.
```suggestion
      let url = transformResult.url || route.url
```

lib/spec/swagger/index.js
```diff
@@ -26,18 +26,20 @@ module.exports = function (opts, cache, routes, Ref, done) {
...
     swaggerObject.paths = {}
     for (const route of routes) {
-      const schema = defOpts.transform
-        ? defOpts.transform(route.schema)
-        : route.schema
+      const transformResult = defOpts.transform
+        ? defOpts.transform(route.schema, route.url)
+        : {}
 
+      const schema = transformResult.schema ?? route.schema
```
climba03003, line 33:
```suggestion
      const schema = transformResult.schema || route.schema
```

README.md
climba03003: I would use a single object for the parameter.
`{ schema, url } => { schema, url }`

climba03003 (changes requested):

README.md
samzilverberg: done, but now its even a bigger breaking change :) need to release a BIG major

climba03003: It is OK, the current syntax is much extendable in future.

index.d.ts
```diff
@@ -127,9 +127,9 @@ export interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
    */
   stripBasePath?: boolean;
   /**
-   * Overwrite the route schema
+   * custom function to transform the route's schema and url
    */
-  transform?: Function;
+  transform?: ({schema, url}: {schema: JSONObject, url: string}) => { schema: JSONObject, url: string };
```
climba03003, line 132: The input is not restricted to JSON object. It can be anything as it is input from user.
```suggestion
  transform?: <S = any>({schema, url}: {schema: S, url: string}) => { schema: JSONObject, url: string };
```

climba03003 (changes requested):

zekth (approved):
lgtm

index.d.ts
samzilverberg: done, I didn't commit ur suggestion directly because `JSONObject` was still on the response

samzilverberg:
rebased on latest main branch and squashed. if you don't have any more suggestions feel free to merge.

JUST REMEMBER this introduces breaking changes so better release as major (or maybe minor will do too, just not patch.......)

darkgl0w:
> rebased on latest main branch and squashed. if you don't have any more suggestions feel free to merge.
> 
> JUST REMEMBER this introduces breaking changes so better release as major (or maybe minor will do too, just not patch.......)

It's because the returned schema is a `JSONObject` ^^

README.md
```diff
@@ -225,7 +225,7 @@ An example of using `fastify-swagger` with `static` mode enabled can be found [h
  | staticCSP          | false            | Enable CSP header for static resources.                                                                                   |
  | stripBasePath      | true             | Strips base path from routes in docs.                                                                                     |
  | swagger            | {}               | [Swagger configuration](https://swagger.io/specification/v2/#swaggerObject).                                              |
- | transform          | null             | Transform method for schema.                                                                                              |
+ | transform          | null             | Transform method for the route's schema and url: `({schema: JSONObject, url: string}) => { schema: JSONObject, url: string }`.                                                                                              |
```
darkgl0w, line 228: This needs to be changed according to @climba03003 suggestion.

index.d.ts
```diff
@@ -127,9 +127,9 @@ export interface FastifyDynamicSwaggerOptions extends FastifySwaggerOptions {
    */
   stripBasePath?: boolean;
   /**
-   * Overwrite the route schema
+   * custom function to transform the route's schema and url
    */
-  transform?: Function;
+  transform?: <S = any>({schema, url}: {schema: S, url: string}) => { schema: S, url: string };
```
darkgl0w, line 132:
```suggestion
  transform?: <S = any>({schema, url}: {schema: S, url: string}) => { schema: JSONObject, url: string };
```

@climba03003 I would go with `unknown` instead of `any` for a user input WDYT ?

climba03003: My intention here is not forcing the user to specify a type and they are good to go. If it is `= unknown`, that means the user is force to specify the type, then I would go for remove the `= unknown`.

samzilverberg: S could be `FastifySchema` (by either direct ref or copy paste the type into this lib) :

https://github.com/fastify/fastify/blob/596e7cad346fa31171f26260a9612b20f4646bcc/types/schema.d.ts#L9-L15

is it a good fit?

README.md
```diff
@@ -241,28 +241,49 @@ If you set `exposeRoute` to `true` the plugin will expose the documentation with
 | `'/documentation/*'`    | External files that you may use in `$ref`  |
 
 <a name="register.options.transform"></a>
-#### Transforms
+#### Transform
 
-To use different schemas such as [Joi](https://github.com/hapijs/joi) you can pass a synchronous `transform` method in the options to convert them back to standard JSON schemas expected by this plugin to generate the documentation (`dynamic` mode only).
+By passing a synchronous `transform` function you can modify the route's url and schema.
+
+some possible uses of this are:
```
Eomm, line 248:
```suggestion
Some possible uses of this are:
```

README.md
```diff
@@ -241,28 +241,49 @@ If you set `exposeRoute` to `true` the plugin will expose the documentation with
...
+
+- add the `hide` flag on schema according to your own logic based on url & schema
+- altering the route url into something that's more suitable for the api spec
+- using different schemas such as [Joi](https://github.com/hapijs/joi) and transforming them to standard JSON schemas expected by this plugin
+
+this option is available in `dynamic` mode only.
```
Eomm, line 254:
```suggestion
This option is available in `dynamic` mode only.
```

README.md
```diff
@@ -241,28 +241,49 @@ If you set `exposeRoute` to `true` the plugin will expose the documentation with
...
+
+example of all the possible uses mentioned:
```
Eomm, line 256:
```suggestion
Examples of all the possible uses mentioned:
```

Eomm (approved):
Just some nit

index.d.ts
climba03003: Yes, it is a possible change. 
```
<S extends FastifySchema = FastifySchema>({schema, url}: {schema: S, url: string}) => { schema: JSONObject, url: string }
```

darkgl0w: This one is better IMO. ^^

samzilverberg: done and it passes the tests 🥳 

I also squashed the commits and force pushed again to make history a lil nicer.

README.md
```diff
@@ -241,30 +241,54 @@ If you set `exposeRoute` to `true` the plugin will expose the documentation with
...
+      headers = undefined,
+      response = undefined,
+      ...rest
     } = schema
-    const transformed = { ...others }
-    if (params) transformed.params = convert(params)
-    if (body) transformed.body = convert(body)
-    if (querystring) transformed.querystring = convert(querystring)
-    return transformed
+    const transformedSchema = { ...rest }
```
darkgl0w, line 272:
```suggestion
    const transformedSchema = Object.assign({}, rest)
```
Replacing rest by something better suited (see my previous comment).

README.md
```diff
@@ -241,30 +241,54 @@ If you set `exposeRoute` to `true` the plugin will expose the documentation with
...
-  transform: schema => {
+  transform: ({ schema, url }) => {
     const {
       params = undefined,
       body = undefined,
       querystring = undefined,
-      ...others
+      headers = undefined,
+      response = undefined,
+      ...rest
```
darkgl0w, line 270: No need to assign undefined as default fallback values here, it is automatically done when using object destructuring.

darkgl0w (commented):
A few little nitpicks.

Other than that it seems good to me :)

README.md
darkgl0w: > why the Object.assign on target empty object? which is essesntialy a shallow copy over if i understand it correctly.

Here `Object.assign({}, rest)` is equivalent to `{ ...rest }` (it is just more explicit) ^^

> we're not mutating anything anywhere, should be ok to just use the original ...rest, or i'm missing something?

By essence `transform` will mutate the schema (and it's what it does in this example), this is why we initialize a new object instead of just doing something like this:

```js
// ...
const { params, body, querystring, headers, response, ...transformedSchema } = schema
let transformedUrl = url

if (params) transformedSchema.params = convert(params)
// ...
```

climba03003 (approved):
LGTM.

README.md
samzilverberg: hmm, this makes me think of 2 things:

1) if we want to ensure no mutation of existing `route.schema` which was passed in, then probably it'd be better to pass `lodash.cloneDeep(route.schema)` to the transform fn.

2)  returned schema should maybe also be some subtype of fastify schema.
to avoid users returning just part of the schema by mistake (without the `rest`).
```
transform?: <S extends FastifySchema = FastifySchema, T extends FastifySchema = FastifySchema>({schema, url}: {schema: S, url: string}) => { schema: T, url: string };
```

WDYT?

darkgl0w: No need to add a new dependency.
It's up to the user to decide if he wants to mutate the original schema or not (`transform` is about letting users apply the mutations of their choice after all 😄). In the documentation we just need to promote good practices IMO and by creating a new object with either `{ ...rest }` or `Object.assign({}, rest)` we do so and it is enough. 😛 

And about the second part, we expect a json schema as the output so checking that it is a json Object is enough and once again it's up to the user to make sure his transform function works as it should 😜

PR #569, opened 2022-03-23
fix: respect patternProperties inner schema

### LINKED ISSUE #548: Type.Record doesn't generate correct schema
### Prerequisites

### Fastify version

3.27.1

### Plugin version

4.15.0

### Node.js version

16.13.1

### Operating system

macOS

11.6.4

### Description

When using @sinclair/typebox and using the Type.Record type, the schema is not represented correctly in the swagger output.

Type.Record lets you type an object where the key of the object is dynamic. In our case the key can be anything, but the value of the key has a specific format.

### Steps to Reproduce

 Using the follow type:

```js
export const testBody = Type.Object({
  foo: Type.Record(
    Type.String(),
    Type.Array(
      Type.Object({
        name: Type.String(),
      })
    )
  ),
});

type Body = Static<typeof testBody>;
```

And this route:

```js
server.route({
  method: "POST",
  url: "/foo",
  async handler(_request, reply) {
    reply.code(204).send();
  },
  schema: { body: testBody },
});
```

The swagger schema example is as follows:

```json
{
  "foo": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  }
}
```

The "string" used here seems to be the type of the key provided to Type.Record.

The typescript type version of the Typebox type works properly. Also it seems that fastify validates the schema correctly.

A post with:

```json
{
  "foo": {
    "bar": "test"
  }
}
```

Returns a proper error from fastify:

```json
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "body.foo['bar'] should be array"
}
```

Closer to correct post continues to validate in fastify:

```json
{
  "foo": {
    "bar": [{}]
  }
}
```

Returns

```json
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "body.foo['bar'][0] should have required property 'name'"
}
```

Using a more simple example `Type.Record` produces the same schema in swagger.

```js
export const testBody = Type.Object({
  foo: Type.Record(Type.String(), Type.Number()),
});
```

Produces:

```json
{
  "foo": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  }
}
```

### Expected Behavior

The swagger schema should be something like:

```
{
  "foo": {
    "additionalProp1": [
      {
        "name": "string"
      }
    ],
  }
}
```

m-ronchi:
fixes #548 

#### Checklist

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -82,7 +82,7 @@ function transformDefsToComponents (jsonSchema) {
   if (typeof jsonSchema === 'object' && jsonSchema !== null) {
     // Handle patternProperties, that is not part of OpenAPI definitions
     if (jsonSchema.patternProperties) {
-      jsonSchema.additionalProperties = { type: 'string' }
+      jsonSchema.additionalProperties = Object.values(jsonSchema.patternProperties)[0];
       delete jsonSchema.patternProperties
     } else if (jsonSchema.const) {
       // OAS 3.1 supports `const` but it is not supported by `swagger-ui`
--- a/test/spec/openapi/schema.js
+++ b/test/spec/openapi/schema.js
@@ -474,6 +474,71 @@ test('support "patternProperties" parameter', async t => {
   })
 })
 
+test('properly support "patternProperties" parameter', async t => {
+  const opt = {
+    schema: {
+      response: {
+        200: {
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #644, opened 2022-08-02
ci: enable license checking

Fdawgs:
See https://github.com/fastify/workflows/issues/49

#### Checklist

- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #646, opened 2022-08-03
replace js-yaml with yaml

Uzlopak:
https://github.com/fastify/fastify-swagger/pull/644#issuecomment-1202966258

I did not run any benchmarks. And it seems that the strings were using double quotes and Dates are outputted without quotes at all, which seems to be correct.

#### Checklist

### DIFF
--- a/lib/mode/static.js
+++ b/lib/mode/static.js
@@ -2,7 +2,7 @@
 
 const path = require('path')
 const fs = require('fs')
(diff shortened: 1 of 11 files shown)


climba03003 (approved):
LGTM.

Benchmark is not important for the document generation.
It only allowed to generate once only, and return by cache after all.

Eomm (approved):

Issue #648, opened 2022-08-06
`exposeHeadRoute` should include HEAD routes 


buschco:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

setting [`exposeHeadRoute`](https://github.com/fastify/fastify/pull/2700) to true in a fastify route, I expect this route to be included with the original Mehtod and the `HEAD` method.

### Motivation

While adding [`exposeHeadRoutes`](https://github.com/supabase/storage-api/pull/171#issuecomment-1206019386) in [supapase-storage-api](https://github.com/supabase/storage-api/) we discovered that the new `HEAD` routes where not exported.

### Example

```
fastify.get<getObjectRequestInterface>(
    '/object/:bucketName/*',
    {
      exposeHeadRoute: true,
```
-> 
```
"/object/{bucketName}/{wildcard}": {
            "get": { ... },
            "head": {  /* same as get but no body */  }
```

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

buschco:
@mcollina I would love to! Can you pre someone else give me a hint where to start? I was looking into the sources and could not find a good point 🧐

mcollina:
I'm not sure either, I would start  writing a reproducible example /unit test and then work from there. Check out the `onRoute` hook, it should be triggered for both GET and HEAD routes.

mcollina:
https://github.com/fastify/fastify-swagger/blob/99a90d425d9a7f903fd7c94eead19d27df060eed/lib/util/common.js#L18 < we are skipping HEAD here.

buschco:
Here is the PR -> https://github.com/fastify/fastify-swagger/pull/649

I think I understand now why they where skipped. I tried to improve this, by only skipping routes that would cause a validation problem by using the same `operationId` multiple times.

see https://swagger.io/docs/specification/paths-and-operations somewhere under `operationId` 
> operationId is an optional unique string used to identify an operation. If provided, these IDs must be unique among all operations described in your API.


--- 

Imo fastify should not use the same `operationId` for the automatic generated routes (e.g. `${operationId}-head` could be a fix). But also the `HEAD` route seems to be the same "operation" of the `GET` but the returned payload is just omited 🤷

PR #649, opened 2022-08-07
feat: support head routes

buschco:
#### Checklist

- [x] run `npm run test`
- [ ]  and `npm run benchmark` (`npm ERR! Missing script: "benchmark"`)
- [ ] documentation is changed or added (tell me where it is necessary)

### DIFF
--- a/lib/util/common.js
+++ b/lib/util/common.js
@@ -11,11 +11,35 @@ function addHook (fastify, pluginOptions) {
   const routes = []
   const sharedSchemasMap = new Map()
 
+  const globalExposeHeadRoutes = fastify.initialConfig.exposeHeadRoutes
+
   fastify.addHook('onRoute', (routeOptions) => {
-    // we need to skip HEAD route
-    // since it may automatically added by fastify
-    // and will conflict with the existing route
-    if (routeOptions.method === 'HEAD') return
+    // fastify will generate HEAD routes if either the global `exposeHeadRoutes`
+    // (default behavior since https://github.com/fastify/fastify/pull/2826) or
+    // `exposeHeadRoute` on route level flag is `true`. If two routes with
+    // operationId are added to the swagger object, it is no longer valid.
+    // therefore we suffix the operationId with `-head`.
+    const hasRouteExposeHeadRouteFlag = routeOptions.exposeHeadRoute != null
+    const exposesHeads = hasRouteExposeHeadRouteFlag
+      ? routeOptions.exposeHeadRoute
+      : globalExposeHeadRoutes
+
+    if (
+      routeOptions.method === 'HEAD' &&
+      exposesHeads === true &&
+      routeOptions.schema !== undefined &&
+      routeOptions.schema.operationId !== undefined
+    ) {
+      routes.push(
+        Object.assign({}, routeOptions, {
+          schema: Object.assign({}, routeOptions.schema, {
+            operationId: `${routeOptions.schema.operationId}-head`
+          })
+        })
+      )
+      return
+    }
+
     routes.push(routeOptions)
   })
 
--- a/test/spec/swagger/schema.js
+++ b/test/spec/swagger/schema.js
@@ -481,3 +481,78 @@ test('support "const" keyword', async t => {
     }
   })
 })
+
+test('support "exposeHeadRoute" option', async (t) => {
+  const fastify = Fastify({ exposeHeadRoutes: false })
+  await fastify.register(fastifySwagger, {
+    routePrefix: '/docs',
+    exposeRoute: true
+  })
+
+  fastify.get('/with-head', {
+    exposeHeadRoute: true,
+    schema: {
+      operationId: 'with-head',
+      response: {
+        200: {
+          description: 'Expected Response',
+          type: 'object',
+          properties: {
+            foo: { type: 'string' }
+          }
+        }
+      }
+    }
+  }, () => {})
+
+  await fastify.ready()
+
+  const swaggerObject = fastify.swagger()
+  const api = await Swagger.validate(swaggerObject)
+
+  t.same(
+    api.paths['/with-head'].get.responses['200'].description,
+    'Expected Response'
+  )
+  t.same(
+    api.paths['/with-head'].head.responses['200'].description,
+    'Expected Response'
+  )
+})
+
+test('support "exposeHeadRoutes" option', async (t) => {
+  const fastify = Fastify({ exposeHeadRoutes: true })
+  await fastify.register(fastifySwagger, {
+    routePrefix: '/docs',
+    exposeRoute: true
+  })
+
+  fastify.get('/with-head', {
+    schema: {
+      operationId: 'with-head',
+      response: {
+        200: {
+          description: 'Expected Response',
+          type: 'object',
+          properties: {
+            foo: { type: 'string' }
+          }
+        }
+      }
+    }
+  }, () => {})
+
+  await fastify.ready()
+
+  const swaggerObject = fastify.swagger()
+  const api = await Swagger.validate(swaggerObject)
+
+  t.same(
+    api.paths['/with-head'].get.responses['200'].description,
+    'Expected Response'
+  )
+  t.same(
+    api.paths['/with-head'].head.responses['200'].description,
+    'Expected Response'
+  )
+})


mcollina (approved):
lgtm

Uzlopak:
Run `npm run lint` and fix the linting issues.

lib/util/common.js
```diff
@@ -12,10 +12,23 @@ function addHook (fastify, pluginOptions) {
...
+    if (
+      routeOptions.method === "HEAD" &&
+      routes.find(
+        ({  schema }) =>
+          (schema != null &&
+            schema.operationId != null &&
+            routeOptions.schema != null &&
+            routeOptions.schema.operationId != null &&
+            schema.operationId === routeOptions.schema.operationId) 
+      ) != null
```
Uzlopak, line 28: Can we not use !== ?

test/spec/swagger/schema.js
```diff
@@ -525,3 +525,130 @@ test('support "const" keyword', async t => {
...
+  t.same(
+    api.paths["/with-head"].get.responses["200"].description,
+    "Expected Response"
+  );
+  t.same(
+    api.paths["/with-head"].head.responses["200"].description,
+    "Expected Response"
+  );
+
+  t.ok(api.paths["/without-head"].head == null);
```
Uzlopak, line 638: Can we not use ===?

test/spec/swagger/schema.js
```diff
@@ -525,3 +525,130 @@ test('support "const" keyword', async t => {
...
+
```
Uzlopak, line 576: Can we not use ===?

Uzlopak (changes requested):
Please use tripple equals and fix the linting issues

lib/util/common.js
buschco: Ah yes of course, but that means also I have to use `undefined` instead of `null`. Are you fine with `!find(...)`?

test/spec/swagger/schema.js
buschco: 👍

test/spec/swagger/schema.js
buschco: 👍

buschco:
@Uzlopak could you please have another look?

lib/util/common.js
```diff
@@ -12,10 +12,23 @@ function addHook (fastify, pluginOptions) {
...
-    // since it may automatically added by fastify
-    // and will conflict with the existing route
-    if (routeOptions.method === 'HEAD') return
+    // we need to skip the HEAD route if there is
+    // already one route with the same operationId.
+    // since https://github.com/fastify/fastify/pull/2826 head routes
+    // are created by default in fastify
+    if (
+      routeOptions.method === 'HEAD' &&
+      routes.find(
```
climba03003, line 21: It depends on the route registration order which will be wrong in future.
I would just add a option to force expose head route here.

It can be placed inside `schema` using `x-` extension or `route options` with a proper name.

climba03003 (changes requested):

lib/util/common.js
buschco: So like this:
```
routeOptions.method === 'HEAD' && routeOptions.schema['x-force-include-head'] !== undefined
```
or so
```
routeOptions.method === 'HEAD' && routeOptions.forceIncludeHead === true
```
?

The second one would require changes in fastify, right?

climba03003: > The second one would require changes in fastify, right?

No, it do not require changes in `fastify`.

buschco: Ah nice 
The second one seems cleaner imo.
Please tell me which do you prefer and if the name `forceIncludeHead` suits you then I change the PR 🙂

lib/util/common.js
```diff
@@ -49,10 +62,12 @@ function addHook (fastify, pluginOptions) {
...
-      return Ref(Object.assign(
-        { applicationUri: 'todo.com' },
-        pluginOptions.refResolver,
-        { clone: true, externalSchemas })
+      return Ref(
+        Object.assign(
+          { applicationUri: 'todo.com' },
+          pluginOptions.refResolver,
+          { clone: true, externalSchemas }
+        )
```
Fdawgs, line 70: Did you leave something like Prettier on? Style has changed but not the content.

buschco: Yes, I prettiered, then the `npm run lint:fix` changed it to this.

I no manually pick my changes to avoid this 👍

Uzlopak:
@Fdawgs 
I think @climba03003  has a point. It should be maybe implemented differently.

mcollina:
Any update on this one? cc @climba03003

climba03003:
> Any update on this one?

It's not updated based on my comment.
The current approach is error-prone and not good for maintenance.

buschco:
sorry i was on vacation 
I will submit a updated version in the coming week 🤞

buschco:
So I just pushed a new approach: As I described here https://github.com/fastify/fastify-swagger/issues/648#issuecomment-1207410705 the problem is a duplicate `operationId`. My solution now, suffixes `HEAD` routes that have an `operationId` while any of `exposeHeadRoutes` or `exposeHeadRoute` is `true`. Also if the flag is set, and the user trys to add a custom `HEAD` route any way, `fastify` already warns the user: https://github.com/fastify/fastify/blob/main/lib/route.js#L358-L360

mcollina (approved):
Lgtm

Uzlopak:
@climba03003 PTAL

Fdawgs (approved):

mcollina:
@Uzlopak @climba03003 ptal

Uzlopak (approved):

lib/util/common.js
```diff
@@ -11,11 +11,35 @@ function addHook (fastify, pluginOptions) {
...
+
   fastify.addHook('onRoute', (routeOptions) => {
-    // we need to skip HEAD route
-    // since it may automatically added by fastify
-    // and will conflict with the existing route
-    if (routeOptions.method === 'HEAD') return
+    // fastify will generate HEAD routes if either the global `exposeHeadRoutes`
+    // (default behavior since https://github.com/fastify/fastify/pull/2826) or
+    // `exposeHeadRoute` on route level flag is `true`. If two routes with
+    // operationId are added to the swagger object, is is no longer valid.
```
Eomm, line 20:
```suggestion
    // operationId are added to the swagger object, it is no longer valid.
```

lib/util/common.js
```diff
@@ -11,11 +11,35 @@ function addHook (fastify, pluginOptions) {
...
+    const hasRouteExposeHeadRouteFlag = routeOptions.exposeHeadRoute != null
+    const exposesHeads = hasRouteExposeHeadRouteFlag
+      ? routeOptions.exposeHeadRoute
+      : globalExposeHeadRoutes
+
+    if (
+      routeOptions.method === 'HEAD' &&
+      exposesHeads === true &&
+      routeOptions.schema !== undefined &&
+      routeOptions.schema.operationId !== undefined
```
Eomm, line 31:
```suggestion
      routeOptions.schema?.operationId !== undefined
```

test/spec/swagger/schema.js
```diff
@@ -525,3 +525,130 @@ test('support "const" keyword', async t => {
     }
   })
 })
+
+test('support "exposeHeadRoute" option', async (t) => {
+  const optWithoutHead = {
+    schema: {
+      response: {
```
Eomm, line 532: I would expect the `operationId` property here

test/spec/swagger/schema.js
```diff
@@ -525,3 +525,130 @@ test('support "const" keyword', async t => {
     }
   })
 })
+
+test('support "exposeHeadRoute" option', async (t) => {
```
Eomm, line 529: This test passes without the code changes on `common.js` file

Eomm (changes requested):

lib/util/common.js
buschco: nice catch 👍

lib/util/common.js
buschco: eslint wont allow me to use this feature, also it was never used in the repo before.

test/spec/swagger/schema.js
buschco: I rewrote the tests, can you please have another look?

test/spec/swagger/schema.js
buschco: I rewrote the tests, can you please have another look?

climba03003 (approved):

mcollina:
@Eomm PTAL

davidebianchi:
Hi all, any update on this PR? It needs some changes? We also need the openapi of the HEAD methods, which at the moment is totally skipped. Can I help to unblock this in some way?

mcollina:
I think this can land

PR #668, opened 2022-09-15
feat: allow to define route log level

leandroandrade:
The purpose of this PR is to allow to define a specific log level in scenarios when we want to `silent` all logs of the swagger route.

#### Checklist

### DIFF
--- a/lib/mode/dynamic.js
+++ b/lib/mode/dynamic.js
@@ -29,13 +29,15 @@ module.exports = function (fastify, opts, done) {
     const initOAuth = opts.initOAuth || {}
     const staticCSP = opts.staticCSP
     const transformStaticCSP = opts.transformStaticCSP
+    const logLevel = opts.logLevel || 'info'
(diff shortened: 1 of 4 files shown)


lib/mode/dynamic.js
```diff
@@ -29,13 +29,15 @@ module.exports = function (fastify, opts, done) {
     const initOAuth = opts.initOAuth || {}
     const staticCSP = opts.staticCSP
     const transformStaticCSP = opts.transformStaticCSP
+    const logLevel = opts.logLevel || 'info'
```
mcollina, line 32: I would not set a default value for this, let it be managed by fastify itself

mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test?

climba03003 (approved):
LGTM.

I believe the next step after this PR should be group all those parameter into a single object instead of mapping it one by one.

mcollina (approved):
lgtm

PR #676, opened 2022-10-07
feat: support local reference in schema, support definitions keyword 

### LINKED ISSUE #524: Error: token "definitions" does not exist
### Prerequisites

### Fastify version

3.25.3

### Plugin version

4.12.1

### Node.js version

16.13.0

### Operating system

Windows

10

### Description

Upon upgrading from 4.12.0 to ^4.12.1, `components.schemas.def-0` is no longer populated with shared schemas/definitions.
This is leading to `Invalid reference token: definitions` being thrown by Redoc, or `token "definitions" does not exist` by other validators, when attempting to load the schema.

### Steps to Reproduce

This issue is present in [Fdawgs/ydh-myydh-crud-api](https://github.com/Fdawgs/ydh-myydh-crud-api).
The [shared schemas are held in a plugin](https://github.com/Fdawgs/ydh-myydh-crud-api/blob/master/src/plugins/shared-schemas/index.js).

#### 4.12.0 example
In v4.12.0 `components.schemas.def-0` looks like so:

```json
"def-0": {
    "$schema": "http://json-schema.org/draft-07/schema#",
    "definitions": {
        "unauthorized": {
            "type": "object",
            "title": "401 Unauthorized",
            "properties": {
                "statusCode": {
                    "type": "number",
                    "const": 401
                },
                "error": {
                    "type": "string",
                    "const": "Unauthorized"
                },
                "message": {
                    "type": "string",
                    "examples": ["missing authorization header"]
                }
            }
        },
        "notFoundDbResults": {
            "type": "object",
            "title": "404 Not Found Response",
            "properties": {
                "statusCode": {
                    "type": "number",
                    "const": 404
                },
                "error": {
                    "type": "string",
                    "const": "Not Found"
                },
                "message": {
                    "type": "string",
                    "enum": [
                        "Invalid or expired search results",
                        "Record does not exist or has already been deleted",
                        "User not found"
                    ]
                }
            }
        },
        "notAcceptable": {
            "type": "object",
            "title": "406 Not Acceptable Response",
            "properties": {
                "statusCode": {
                    "type": "number",
                    "const": 406
                },
                "error": {
                    "type": "string",
                    "const": "Not Acceptable"
                },
                "message": {
                    "type": "string",
                    "const": "Not Acceptable"
                }
            }
        },
        "tooManyRequests": {
            "type": "object",
            "title": "429 Too Many Requests Response",
            "properties": {
                "statusCode": {
                    "type": "number",
                    "const": 429
                },
                "error": {
                    "type": "string",
                    "const": "Too Many Requests"
                },
                "message": {
                    "type": "string",
                    "examples": [
                        "Rate limit exceeded, retry in 1 minute"
                    ]
                }
            }
        },
        "internalServerError": {
            "type": "object",
            "title": "500 Internal Server Error Response",
            "properties": {
                "statusCode": {
                    "type": "number",
                    "const": 500
                },
                "error": {
                    "type": "string",
                    "const": "Internal Server Error"
                },
                "message": {
                    "type": "string",
                    "examples": [
                        "Unable to delete read receipt from database",
                        "Unable to return result(s) from database",
                        "Unable to update patient preference in database",
                        "Unable to update read receipt in database"
                    ]
                }
            }
        },
        "serviceUnavailable": {
            "type": "object",
            "title": "503 Service Unavailable",
            "properties": {
                "statusCode": {
                    "type": "number",
                    "const": 503
                },
                "code": {
                    "type": "string",
                    "const": "FST_UNDER_PRESSURE"
                },
                "error": {
                    "type": "string",
                    "const": "Service Unavailable"
                },
                "message": {
                    "type": "string",
                    "const": "Service Unavailable"
                }
            }
        }
    },
    "type": "object",
    "title": "Responses",
    "description": "Common response schemas"
},
```

#### ^4.12.1 example
When launching the API with any version of fastify-swagger ^4.12.1, `components.schemas` looks like so:

```jsonc
"schemas": {
    "def-0": {
        "$schema": "http://json-schema.org/draft-07/schema#",
        "type": "object",
        "title": "Responses",
        "description": "Common response schemas"
    },
    "def-1": {
        "type": "object",
        "title": "401 Unauthorized",
        "properties": {
            "statusCode": {
                "type": "number",
                "const": 401
            },
            "error": {
                "type": "string",
                "const": "Unauthorized"
            },
            "message": {
                "type": "string",
                "example": "missing authorization header"
            }
        }
    },
}

```

As you can see above, it seems the definitions have moved outside of the original def-0 object.
The route schemas still refer to non-existent definitions at def-0:

```json
"406": {
    "description": "Not Acceptable",
    "content": {
      "text/plain": {
        "schema": {
          "$ref": "#/components/schemas/def-0/definitions/notAcceptable",
          "description": "Not Acceptable"
        }
      }
    }
  },
```

### Expected Behavior

$refs to refer to correct definitions, either by adding them back to def-0 or by referring to def-*

### LINKED ISSUE #639: JSON Schema Definitions not work
### Prerequisites

### Issue

```
{
  type: 'object',
  definitions: {
    foo: {
      $id: '#address',
      type: 'object',
      properties: {
        city: { type: 'string' }
      }
    }
  }
}
```

and used in my route schema 

```
 schema: {
      body: {
      }
    },
```

### LINKED ISSUE #675: Support for 'definitions' in json-schema
### Prerequisites

### 🚀 Feature Proposal

https://github.com/fastify/fastify-swagger/issues/639#issuecomment-1271364036

> The 'definitions' keyword in the component(3.0)/definition(2.0) item schema itself is not allowed at all, it is allowed only in top-lvl as 'definition' if we are using the OpenAPI 2.0 version. The changes introduced in https://github.com/fastify/fastify-swagger/pull/472 (https://github.com/fastify/fastify-swagger/pull/472/files#diff-3faaf157cccc3cc534ec616affca5b4f5df9817c937b7ada4f110afd2770aae2R35) is deleting the 'definitions' key from the schema item object, therefore preventing any possible errors on the schema validation, but also preventing the correct use of json-schema with this keyword.
> 
> In the case of json-schema coming from fastify, it seems like we have two ways to solve this problem, first one is to mutate 'definitions' keyword to 'properties' as stated in https://github.com/fastify/fastify-swagger/issues/524#issuecomment-1203660976 workaround, but i have a suspicion that this can cause troubles when resolving some deep-nested schema, currently I'm in the process of checking it.
> 
> The second one is to move all nested definitions from 'definitions' obj to top-lvl of components/definitions, but that doesn't seem to follow the intended behaviour from original json-schema(?)

### Motivation

Addressing the issue in #639 

### Example

_No response_

asc11cat:
Addressing #675 and #639

#### Checklist


test/spec/openapi/refs.js
```diff
@@ -282,3 +282,39 @@ test('uses examples if has property required in body', async (t) => {
...
+        SchemaA: {
+          type: 'object',
+          properties: {
+            id: { type: 'string' }
+          }
+        },
+        SchemaB: {
+          type: 'object',
+          properties: {
+            example: { $ref: 'NestedSchema#/definitions/SchemaA' }
```
ivan-tymoshenko, line 303: @asc11cat Will it work with a local $ref `#/definitions/SchemaA'`?

ivan-tymoshenko (commented):
Will it work for definitions in nested schemas?

test/spec/openapi/refs.js
asc11cat: > @asc11cat Will it work with a local $ref `#/definitions/SchemaA'`?

@ivan-tymoshenko, won't work, it seems like we lack support for the relative path overall as I mentioned in https://github.com/fastify/fastify-swagger/issues/675#issuecomment-1272280205

Should we add some more magic with replaces, or can there be a better way?

ivan-tymoshenko: I wrote json schema RefResolver for FJS. I don't know if it work here, because swagger/openapi has some specific rules. You can check if it work here.
https://github.com/fastify/fast-json-stringify/blob/master/lib/ref-resolver.js

ivan-tymoshenko: It has two public methods `addSchema` and `getSchema`. If you have any questions, feel free to ask.

asc11cat:
@ivan-tymoshenko I faced some issues related to relative definitions, and will appreciate if you can take a look

Right now when we will try to do something like this(even on master):
```
instance.addSchema({
      $id: 'NestedSchema',
      properties: {
        SchemaA: {
          type: 'object',
          properties: {
            id: { type: 'string' }
          }
        },
        SchemaB: {
          type: 'object',
          properties: {
            example: { $ref: '#/properties/SchemaA' }
          }
        }
      }
    })
```

We will get the 'Token "properties" does not exist' exception from Swagger.validate. As i understand, the relative definitions is the correct behavior for json-schema(so for json-schema-resolver), but not for a swagger/openapi one, so we should patch the relative definitions to absolute?

ivan-tymoshenko:
> so we should patch the relative definitions to absolute?

According to json schema spec if you define subschema with a $id property the root inside this schema chages.
```js
{
  $id: 'globalsSchema',
  properties: {
    a: { $ref: '#' }, // points to the 'globalSchema'
    b: {
      $id: 'localSchema',
      type: 'object',
      properties: {
        c: { $ref: '#' } // points to the local schema
      }
    }
  }
}
```
Swagger doesn't work like that. As I understand when you merge all schemas to the swagger schema you get one big schema with a new root. So `#` is no longer point to you local schema, but for one global schema. So yes, you should replace all local refs with absolute refs.

asc11cat:
> > so we should patch the relative definitions to absolute?
> 
> According to json schema spec if you define subschema with a $id property the root inside this schema chages.
> 
> ```js
> {
>   $id: 'globalsSchema',
>   properties: {
>     a: { $ref: '#' }, // points to the 'globalSchema'
>     b: {
>       $id: 'localSchema',
>       type: 'object',
>       properties: {
>         c: { $ref: '#' } // points to the local schema
>       }
>     }
>   }
> }
> ```
> 
> Swagger doesn't work like that. As I understand when you merge all schemas to the swagger schema you get one big schema with a new root. So `#` is no longer point to you local schema, but for one global schema. So yes, you should replace all local refs with absolute refs.



> > so we should patch the relative definitions to absolute?
> 
> According to json schema spec if you define subschema with a $id property the root inside this schema chages.
> 
> ```js
> {
>   $id: 'globalsSchema',
>   properties: {
>     a: { $ref: '#' }, // points to the 'globalSchema'
>     b: {
>       $id: 'localSchema',
>       type: 'object',
>       properties: {
>         c: { $ref: '#' } // points to the local schema
>       }
>     }
>   }
> }
> ```
> 
> Swagger doesn't work like that. As I understand when you merge all schemas to the swagger schema you get one big schema with a new root. So `#` is no longer point to you local schema, but for one global schema. So yes, you should replace all local refs with absolute refs.

Got it, ill try to see what i can do here in a few days, i think it will be better to rethink on how this process of schema mutation from json-schema->openapi works altogether, because right now its really difficult to extend it on need.

asc11cat:
@ivan-tymoshenko finally got some time to finish this PR

Now local references are being transformed to absolute, and 'definitions' keyword is being merged with 'properties', with precedence to the last one

ivan-tymoshenko:
I will check it deeply ASAP. But I have a question for now. I think that problem is in the [json-schema-resolver](https://github.com/Eomm/json-schema-resolver). Shouldn't functions like `localSchemaRefToAbs` and `patchDefinitionsKeywordInSchema` be there?

ivan-tymoshenko:
@Eomm Can you help us here?

asc11cat:
> I will check it deeply ASAP. But I have a question for now. I think that problem is in the [json-schema-resolver](https://github.com/Eomm/json-schema-resolver). Shouldn't functions like `localSchemaRefToAbs` and `patchDefinitionsKeywordInSchema` be there?

Both functions are patching behaviour related only to openapi schema standart, not the json-schema one. In the first case, openapi schema don't support the local refs, in the second - no support for 'definitions' keyword

asc11cat:
Now tests should pass, replaceAll is not supported in node 14, so I replaced it

Eomm (dismissed):
Blocking it as I'm reviewing it.

> I think that problem is in the json-schema-resolver. Shouldn't functions like localSchemaRefToAbs and patchDefinitionsKeywordInSchema be there?

At first read of this PR: yes, I think the same.

test/spec/openapi/refs.js
```diff
@@ -282,3 +282,155 @@ test('uses examples if has property required in body', async (t) => {
   t.ok(schema.parameters)
   t.same(schema.parameters[0].in, 'query')
 })
+
+test('support absolute refs in schema', async (t) => {
```
Eomm, line 286: These due tests:

- `support absolute refs in schema`
- `support relative refs in schema`

Are already supported by json-schema-resolver https://github.com/Eomm/json-schema-resolver/pull/8

The `openapi` conversion instead ( test named `support definitions keyword in schema`) is provided by this module.

So the questions are:

- is this an issue with the conversion from swagger to openapi?
- is this an issue with `fast-json-stringify`? (as it resolves the schemas by its own)

So, right now, I think this fix does not solve the core problem

Eomm: Thanks for the detailed information, I need to catch up on this 👍🏽 
I want to explore a bit the openapi repo as it is source of best practices in these cases

lib/util/common.js
```diff
@@ -210,11 +213,76 @@ function resolveSwaggerFunction (opts, cache, routes, Ref, done) {
...
+
+function patchDefinitionsKeywordInSchema (schema) {
+  if (!schema) return schema
+
+  Object.keys(schema).forEach(key => {
+    if (key === '$ref' && schema[key]) {
+      schema[key] = schema[key].split('definitions').join('properties')
+    } else if (schema[key] && typeof schema[key] === 'object') {
+      if (key === 'definitions') {
+        schema.properties = {
```
ivan-tymoshenko, line 265: Why do you move schema from the `definitions` to the `properties`? You shouldn't create new properties.

ivan-tymoshenko (changes requested):

ghost (approved):

Eomm:
To review this PR, I need ~4h and I need to find such time 😮‍💨

mathiasbockwoldt:
Hi @Eomm, did you find some time, yet? Would be great to get this PR merged. 🚀

melroy89:
> To review this PR, I need ~4h and I need to find such time 😮‍💨

This is now blocking for 1 year.. this is not good. Hopefully else can review and merge this PR.

mcollina:
@asc11cat can you refresh this PR? There are quite a few conflicts now.

Eomm (commented):
Stepping back

ivan-tymoshenko (changes requested):
Finally I have some time to review this PR. As I understand it still tries to move json schema `definitions` under some schema `properties`. I don't think it's a clean solution and it probably won't work with a recurcive refs. If swagger/openapi doesn't support the `definitions` keyword I think the best solution would be to move all `definitions` under `#/components/schemas`.

ivan-tymoshenko:
After 6 hours trying to understand what is going on here I have few conclusions:
- the library needs some brushing. There are a lot of bicycles invented here, so when you touching one part of the code another one is failing.
- about the issue. IMHO before adding a json-schema to openapi-swagger schema we should _recursively_ check for all keywords that are restricted to use in openapi/swagger (not only the `definition` keyword). When we find the keyword we need to remove it from the original schema and add it to the global definitions/components section. At the same time we sshould have some global ref-resolver that would resolve any reference including the refs that includes path with restricted kwyword. Probably it should have some internal mapping like that `#/components/schemas/def-0/definitions/foo/properties/bar` -> `#/components/schemas/def-42`

gurgunday:
I agree that this plugin needs a rewrite/refactor

navalex:
Hello, any news on this case ? Does this PR just dropped because too complex too heavy with the current state of the library ? Or can we expect a little fix to manage it ?

Thanks :)

Tony133:
Closing this PR due to inactivity: it has had no updates since 2023 and is no longer aligned with the current codebase (several conflicts with `main`, and the internal APIs it touches have changed since).

The underlying problem is still tracked in #639. If you'd like to pick this up again, please open a new PR based on the current `main`.

Issue #686, opened 2022-10-30
Hide schema in swagger dcoument


imjuni:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Schema hide field add on option.

```ts
const server = fastify();

await server.register(fastifySwagger, {
  openapi: {
    info: {
      title: "Maeum boilerplate",
      description: "Maeum boilerplate Swagger Document",
      version: "0.2.0",
    },
  },
  // add new option
  hideSchemas: [
    "IReqPokeDetailQuerystring",
    "IReqPokeDetailParams",
    "IConfiguration",
    "IServer",
    "IEndpoint",
  ],
});

await server.register(fastifySwaggerUI, swaggerUiConfig());

server.addSchema({ $id: 'IReqPokeDetailQuerystring' /* ... another configuration */ });
server.addSchema({ $id: 'IReqPokeDetailParams' /* ... another configuration */ });
server.addSchema({ $id: 'IConfiguration' /* ... another configuration */ });
server.addSchema({ $id: 'IServer' /* ... another configuration */ });
server.addSchema({ $id: 'IEndpoint' /* ... another configuration */ });

server.get('/pokemon/:name', {
  querystring: { $ref: 'IReqPokeDetailQuerystring' }
  params: { $ref: 'IReqPokeDetailParams' }
}, () => { /** server code */ });
```

### Motivation

I prefer to add and use all schemas on `fastify schema store`. So I meet some problem like below, 

![fastify-swagger-schemas](https://user-images.githubusercontent.com/1166575/198879237-4fed3443-917c-4a19-a44b-df2a6130d156.jpg)

If all schemas are registered in `fastify schema store`, all schemas are exposed to the document as shown in the picture. This problem make to complex document.

So I think that schema `Request Schema`, `Server Config` have to hide for document reader.

Add content at 2022-11-02

#### Case01
```js
server.get('/pokemon/:name', {
  querystring: { type: 'object', properties: { id: { type: 'number' } } },
  params: { type: 'object', properties: { name: { type: 'string' } } }
}, () => { /** server code */ });
```

#### Case02
```js
server.addSchema({ $id: 'IReqPokeDetailQuerystring' /* ... another configuration */ });
server.addSchema({ $id: 'IReqPokeDetailParams' /* ... another configuration */ });

server.get('/pokemon/:name', {
[...]

PR #687, opened 2022-10-31
feat: Add hideSchemas swagger option (#686)

imjuni:
* feat: add hideSchemas swagger option
* test: add tests for hideSchemas swagger option
* docs: add hideSchemas swagger option

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,7 @@ An example of using `@fastify/swagger` with `static` mode enabled can be found [
  | swagger            | {}               | [Swagger configuration](https://swagger.io/specification/v2/#swaggerObject).                                              |
  | transform          | null             | Transform method for the route's schema and url. [documentation](#register.options.transform).                                                                                              |
  | refResolver        | {}               | Option to manage the `$ref`s of your application's schemas. Read the [`$ref` documentation](#register.options.refResolver) |
+ | hideSchemas        | []               | Option to control hiding of schemas in swagger document definitions                                                        |
 
 <a name="register.options.transform"></a>
 #### Transform
--- a/lib/spec/openapi/index.js
+++ b/lib/spec/openapi/index.js
@@ -20,9 +20,17 @@ module.exports = function (opts, cache, routes, Ref, done) {
     const openapiObject = prepareOpenapiObject(defOpts, done)
 
     ref = Ref()
+
+    const definitions = {}
+    for (const [$id, schema] of Object.entries(ref.definitions().definitions)) {
+      if (defOpts.hideSchemas.findIndex(hideSchema => hideSchema === schema.$id) < 0) {
+        definitions[$id] = schema
+      }
+    }
+
     openapiObject.components.schemas = prepareOpenapiSchemas({
       ...openapiObject.components.schemas,
-      ...(ref.definitions().definitions)
+      ...definitions
     }, ref)
 
     const serverUrls = resolveServerUrls(defOpts.servers)
--- a/test/spec/openapi/schema.js
+++ b/test/spec/openapi/schema.js
@@ -849,3 +849,73 @@ test('support query serialization params', async t => {
   t.equal(api.paths['/'].get.parameters[0].style, 'deepObject')
   t.equal(api.paths['/'].get.parameters[0].explode, false)
 })
+
+test('support hide schema', async t => {
+  t.plan(2)
+  const fastify = Fastify()
+  await fastify.register(fastifySwagger, { ...openapiOption, hideSchemas: ['schema-02'] })
+
+  fastify.addSchema({
+    $id: 'schema-01',
+    type: 'object',
+    properties: {
+      id: { type: 'number', description: 'data id' }
+    },
+    required: ['id']
+  })
+
+  fastify.addSchema({
+    $id: 'schema-02',
+    type: 'object',
+    properties: {
+      id: { type: 'number', description: 'data id' }
+    },
+    required: ['id']
+  })
+
+  fastify.addSchema({
+    $id: 'schema-03',
+    type: 'object',
+    properties: {
+      id: { type: 'number', description: 'data id' }
+    },
+    required: ['id']
+  })
+
+  fastify.get('/', {
+    schema: {
+      querystring: { $ref: 'schema-02' },
+      response: {
+        200: {
+          type: 'object'
+        }
+      }
+    }
+  }, () => {})
+
+  await fastify.ready()
+
+  const openapiObject = fastify.swagger()
+  await Swagger.validate(openapiObject)
+
+  const definitions = openapiObject.components.schemas
+  t.ok(definitions)
+  t.same(definitions, {
+    'def-0': {
+      type: 'object',
+      properties: {
+        id: { type: 'number', description: 'data id' }
+      },
+      required: ['id'],
+      title: 'schema-01'
+    },
+    'def-2': {
+      type: 'object',
+      properties: {
+        id: { type: 'number', description: 'data id' }
+      },
+      required: ['id'],
+      title: 'schema-03'
+    }
+  })
+})
--- a/test/spec/swagger/schema.js
+++ b/test/spec/swagger/schema.js
@@ -481,3 +481,73 @@ test('support "const" keyword', async t => {
     }
   })
 })
+
+test('support hide schema', async t => {
+  t.plan(2)
+  const fastify = Fastify()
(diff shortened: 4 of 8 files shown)


test/spec/swagger/schema.js
```diff
@@ -481,3 +481,74 @@ test('support "const" keyword', async t => {
...
+    schema: {
+      response: {
+        200: {
+          type: 'object'
+        }
+      }
+    }
+  }
+
+  fastify.get('/', opts9, () => {})
```
mcollina, line 527: Can you please add a test that uses `schema-02` as a ref?

README.md
```diff
@@ -219,6 +219,7 @@ An example of using `@fastify/swagger` with `static` mode enabled can be found [
  | swagger            | {}               | [Swagger configuration](https://swagger.io/specification/v2/#swaggerObject).                                              |
  | transform          | null             | Transform method for the route's schema and url. [documentation](#register.options.transform).                                                                                              |
  | refResolver        | {}               | Option to manage the `$ref`s of your application's schemas. Read the [`$ref` documentation](#register.options.refResolver) |
+ | hideSchemas        | []               | Option to control hiding of schemas.                                                                                       |
```
Uzlopak, line 222: I dont get it. Hiding of Schemas from who or where?

test/spec/swagger/schema.js
imjuni: Thank you for your review.

I change code like that,

```ts
fastify.get('/', {
    schema: {
      querystring: { $ref: 'schema-02' },
      response: {
        200: {
          type: 'object'
        }
      }
    }
  }, () => {})
```

README.md
imjuni: Thank you for your review,

Change description like that,

```
Option to control hiding of schemas in swagger document definitions
```

How about this? I mean only hide that from definition sections in swagger documents.

test/spec/openapi/schema.js
```diff
@@ -849,3 +849,73 @@ test('support query serialization params', async t => {
...
+          type: 'object'
+        }
+      }
+    }
+  }, () => {})
+
+  await fastify.ready()
+
+  const openapiObject = fastify.swagger()
+  await Swagger.validate(openapiObject)
```
mcollina, line 899: I don't understand how this would be valid.

climba03003: `Swagger.validate` only validate `structure`. It does not validate the linkage.
Yes, if `schema-02` is hidden, it should not be a valid one.

mcollina (approved):
lgtm

test/spec/openapi/schema.js
imjuni: I think that difference between json schema store and swagger document definition. Validation need every schema but swagger document don't have to display every definitions.

Also, I can move the logic of hiding the schema to the last time of the document building like OpenAPI v2.

Uzlopak:
I am now too lazy to read the code. But the only way to avoid those schemas to be rendered in the final document should be by resolving the references with the schemas.

lib/spec/openapi/index.js
```diff
@@ -20,9 +20,12 @@ module.exports = function (opts, cache, routes, Ref, done) {
     const openapiObject = prepareOpenapiObject(defOpts, done)
 
     ref = Ref()
+    const definitions = Object.entries(ref.definitions().definitions)
+      .filter(([, schema]) => defOpts.hideSchemas.findIndex(hideSchema => hideSchema === schema.$id) < 0)
+      .reduce((defs, [$id, schema]) => ({ ...defs, [$id]: schema }), {})
```
climba03003, line 25: Can you change this code into a single `for loop`?
It repeatedly cloning the `defs`.

climba03003 (changes requested):

lib/spec/openapi/index.js
imjuni: Sure, I chaged it.

climba03003 (approved):

imjuni:
@Uzlopak , @mcollina Here is example, 

```js
const fastify = require('fastify');
const server = fastify();

// case 01. error raise when server start
server.addSchema({
  $id: 'IReqPokeDetailQuerystring',
  type: 'object',
  properties: {
    tid: {
      $ref: 'missing reference!',
    },
  },
  required: ['tid'],
});

// case 02. error raise when request call
server.get(
  '/pokemon/:name',
  {
    schema: {
      querystring: {
        type: 'object',
        properties: {
          tid: {
            $ref: 'missing reference!',
          },
        },
        required: ['tid'],
      },
    },
  },
  () => {
    /* ... */
  },
);
```

I prefer what case01 then case02. Because more than quickly aware error from validation schema. But case01 generate so many DTO in swagger document. My company project already over 100 DTOs. Since Querystirng, Headers, Params, and Body each generate a DTO, the DTO can be increased very much. So I hope hide Querystring, Headers, Params and Body DTO. Because Querystring, Headers, Params and Body already describe in API spec.

Uzlopak:
I understand that, but you need to resolve the references. If the DTO is not used anyway, it should not be added to the final swagger document.

But you are just manually suppress the propagation to the final document.

So e.g. i would actually expect that if you specifiy IReqPokeDetailQuerystring as a reference that if it is used somewhere to be resolved so that the specification is still valid and remove it from the final swagger documentation.  If it is not referenced anywhere, it would be anyway removed. 

So instead of hideSchemas i would expect a resolveReferences option, where you pass an Array or maybe  a function where the name of a reference is passed and then decided if it should be resolved or not

```typescript
resolveReference: (name: string) => boolean
```

Then you could do something like

```js
function resolveReference(name) {
   return name.endsWith("Querystring") === false
}
```

imjuni:
I understand resolve reference. But some difference of this case. First I try to remove reference like that,

```js
ref = Ref()
const prevDefinitions = prepareSwaggerDefinitions({
  ...swaggerObject.definitions,
  ...(ref.definitions().definitions)
}, ref)

// remove reference
const definitions = {}
for (const [$id, schema] of Object.entries(prevDefinitions)) {
  if (defOpts.resolveReference(schema.$id)) {
    definitions[$id] = schema
  }
}
```

But this approach fail to build document in case of OpenAPI v2. Because OpenAPI v2 use reference for document building. This work raise error from [prepareSwaggerMethod](https://github.com/fastify/fastify-swagger/blob/c8d675d819f7fd274f00a20144a4efdca0c55778/lib/spec/swagger/index.js#L46). But OpenAPI v3 fine work this approach. I thought it would be better for v2 and v3 to behave the same. Also I can chage option type `string[]` to `Function`. Like that, 

```js
const definitions = {}
for (const [$id, schema] of Object.entries(swaggerObject.definitions)) {
  if (defOpts.isHideReference(schema.title)) {
    definitions[$id] = schema
  }
}
```

Uzlopak:
Imho you are just removing the DTOs from the final document.  I mean your unit tests actually show it beatifully. 
So despite you added schema-0, schema-1 and schema-2 you manually remove schema-1, and because you dont use it in your schema building the schema succeeds . If we would remove schema-2, we should get a FST_ERR_SCH_SERIALIZATION_BUILD-Error.

We should actually fist add a flag which removes unused definitions from the final swagger-documentation.

Example unit test:

```js
test('remove unused schemas', async t => {
  t.plan(2)
  const fastify = Fastify()
  await fastify.register(fastifySwagger, { ...openapiOption, removeUnusedSchemas: true })

  fastify.addSchema({
    $id: 'schema-01',
    type: 'object',
    properties: {
      id: { type: 'number', description: 'data id' }
    },
    required: ['id']
  })

  fastify.addSchema({
    $id: 'schema-02',
    type: 'object',
    properties: {
      id: { type: 'number', description: 'data id' }
    },
    required: ['id']
  })

  fastify.addSchema({
    $id: 'schema-03',
    type: 'object',
    properties: {
      id: { type: 'number', description: 'data id' }
    },
    required: ['id']
  })

  fastify.get('/', {
    schema: {
      querystring: { $ref: 'schema-02' },
      response: {
        200: {
          type: 'object'
        }
      }
    }
  }, () => {})

  await fastify.ready()

  const openapiObject = fastify.swagger()
  await Swagger.validate(openapiObject)

  const definitions = openapiObject.components.schemas
  t.ok(definitions)
  t.same(definitions, {
    'def-2': {
      type: 'object',
      properties: {
        id: { type: 'number', description: 'data id' }
      },
      required: ['id'],
      title: 'schema-03'
    }
  })
})
```

In a second step, we should implement a flag, which forces to resolve specified DTOs. And then remove them from being exposed in the final swagger document.

Issue #691, opened 2022-11-13
Cannot pass definitions field when using openapi configuration


imjuni:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.9.2

### Plugin version

8.1.0

### Node.js version

14.20.1

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Ubuntu 20.04.3 LTS

## Issue 
Cannot found $ref schema when custom SchemaController apply on fastify server instance. `@fastify/swagger` display error message like that,

```json
{
  "statusCode": 500,
  "error": "Internal Server Error",
  "message": "Cannot read property 'type' of undefined"
}
```

### [Custom SchemaController](https://github.com/imjuni/maeum/blob/06ffd5a6f54fa4cd5ca7055d2db4d967e97f2cf2/src/server/module/optionFactory.ts#L31)
ajv schema store split from fastify server instance because schema store use other case. eg. redis, rabbitmq pub/sub or file upload etc(share $ref schema). 

### Analysis
1. I believe the following code is a bug.
    * [build querystring](https://github.com/fastify/fastify-swagger/blob/d504587aa414e65ba7ac6ea2774595fb9d507f51/lib/spec/openapi/utils.js#L407), [build params](https://github.com/fastify/fastify-swagger/blob/d504587aa414e65ba7ac6ea2774595fb9d507f51/lib/spec/openapi/utils.js#L412), [build headers](https://github.com/fastify/fastify-swagger/blob/d504587aa414e65ba7ac6ea2774595fb9d507f51/lib/spec/openapi/utils.js#L413), [build cookies](https://github.com/fastify/fastify-swagger/blob/d504587aa414e65ba7ac6ea2774595fb9d507f51/lib/spec/openapi/utils.js#L416)
    * `definitions` field remove from the option so that change `openapiObject.definitions` to `openapiObject.components.schemas`
    * `definitions` is always `undefined` so no way to pass externalSchemas
```js
// as-is
if (schema.querystring) resolveCommonParams('query', parameters, schema.querystring, ref, openapiObject.definitions, securityIgnores.query)
if (schema.body) {
  openapiMethod.requestBody = { content: {} }
  resolveBodyParams(openapiMethod.requestBody, schema.body, schema.consumes, ref)
}
if (schema.params) resolveCommonParams('path', parameters, schema.params, ref, openapiObject.definitions)
[...]

PR #692, opened 2022-11-16
fix: externalSchemas problem when using openapi (#691)

imjuni:
* fix cannot pass externalSchemas when using openapi configuration

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "fastify": "^4.0.0",
     "fluent-json-schema": "^4.0.0",
     "joi": "^17.6.0",
-    "joi-to-json": "^2.2.4",
+    "joi-to-json": "^3.1.1",
     "qs": "^6.10.3",
     "standard": "^17.0.0",
     "tap": "^16.2.0",


imjuni:
@Uzlopak Hi, I add new testcase `external schema reference`.

mcollina (approved):
lgtm

imjuni:
is any progress?

test/spec/openapi/refs.js
```diff
@@ -282,3 +282,46 @@ test('uses examples if has property required in body', async (t) => {
...
+          }
+        }
+      }
+    }
+  })
+
+  fastify.get('/', {
+    schema: {
+      response: {
+        200: {
```
Uzlopak, line 310: You dont test your change imho. here we need something like { $ref: 'external-schema-01' }

Uzlopak (changes requested):
I am quite hesitant. The PR feels "wrong".
Shouldnt it be also working by using addSchema? Why is there  not a test, that with addSchema also components.schema is set with schemas from addSchema?

Also shouldnt be the test using the "external" schema as reference to prove that it is used internally by this plugin and not only existing in the final openapi definition?

Issue #625, opened 2022-06-26
OAS 3.1: Examples in the open api spec schema property should be an array


avin-kavish:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4

### Plugin version

7.3

### Node.js version

16

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

20

### Description

`schema` object in Open API spec follows the JSON Schema standard as mentioned [here](https://spec.openapis.org/oas/v3.1.0#schema-object).

`example` field is deprecated as mentioned [here](https://spec.openapis.org/oas/v3.1.0#fixed-fields-19).

`examples` field is an array as mentioned [here](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-examples) in the JSON schema spec. 

But, fastify-swagger generates the examples field as an object using this method `convertExamplesArrayToObject`, which breaks some implementations of Open API spec viewers.

Also, note, not every examples field is supposed to be an array, only when JSON schema is expected. For example, the examples of the response object are indeed an object, as mentioned [here](https://spec.openapis.org/oas/v3.1.0#media-type-object).

### Steps to Reproduce

Generate open api spec when you have validation schema like,

```ts
import { Type as t } from '@sinclair/typebox'

t.String({
  examples: [ 'haha', 'haha2' ]
})
```

generates schema like,

```
schema:
  type: string
  examples:
    haha:
      value: haha
    haha2:
      value: haha2
```

### Expected Behavior

```
schema:
  type: string
  examples:
      - haha
      - haha2
```
I think the solution would need to conditionally generate an example array in JSON Schema and an example object in other places.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

avin-kavish:
Looks like there were some changes in #616 related to this.

mcollina:
cc @mxck

mcollina:
All of you are significantly more expert than me on the OpenAPI spec. I defer to contributors on these OAS features

mxck:
@avin-kavish doesn't swagger already support 3.1? 
I redesigned the examples field in the past PR for version 3.0

avin-kavish:
> doesn't swagger already support 3.1?

I'm not sure what you mean. Do you mean this spec? https://swagger.io/specification/

mxck:
[...]

PR #702, opened 2022-12-09
fix `examples` in json schema

trim21:
#### Checklist

fix https://github.com/fastify/fastify-swagger/issues/625

### DIFF
--- a/README.md
+++ b/README.md
@@ -780,7 +780,7 @@ Will generate this in the OpenAPI v3 schema's `path`:
                   "foo": { "type": "string" },
                   "bar": { "type": "string" }
                 },
-                "example": { "foo": "bar", "bar": "baz" }
+                "examples": [{ "foo": "bar", "bar": "baz" }]
               }
             }
           },
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -107,7 +107,7 @@ function resolveServerUrls (servers) {
   return resolvedUrls
 }
 
-function transformDefsToComponents (jsonSchema, inSchema = false) {
+function transformDefsToComponents (jsonSchema) {
   if (typeof jsonSchema === 'object' && jsonSchema !== null) {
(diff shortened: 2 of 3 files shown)


mcollina:
I think we still need to support OpenAPI 3.0 too.

trim21:
> I think we still need to support OpenAPI 3.0 too.

I don't think this will break openapi 3.0, it just didn't replace `examples` property in json schema object

climba03003 (changes requested):
I would move the `examples` mutation to the proper place that should happen.

For example, [resolveResponse](https://github.com/fastify/fastify-swagger/blob/7546f2311c677a0a06acf112c0ebc8a50787e48d/lib/spec/openapi/utils.js#L318).
It can ensure, we are properly mutate the top-level only.

trim21:
> I would move the `examples` mutation to the proper place that should happen.
> 
> For example, [resolveResponse](https://github.com/fastify/fastify-swagger/blob/7546f2311c677a0a06acf112c0ebc8a50787e48d/lib/spec/openapi/utils.js#L318). It can ensure, we are properly mutate the top-level only.

`transformDefsToComponents` is called on path, params, response and query, this will require to write it multiple times

trim21:
I guess you want to move `examples` transform in `schemaToMedia`?

climba03003:
Actually, `cookie`, `query`, `path`, `header` already doing it's own transform for the `properties.examples`. There are no need to mutate the `examples` for them.

For `body`, it reuse the same logic inside `schemaToMedia` which does require some mutation.

For `response`, it is the only one need to do more things.

trim21:
I think this is what you are expecting?

climba03003 (approved):

Eomm (approved):

trim21:
any releasing plan？ I'm currently encountering a openapi rendering issue will be fixed by this PR.

Eomm:
Done in 8.2.1

Issue #701, opened 2022-12-06
Header enum value is not getting generated as available values on @fastify/swagger


vishnutmohan:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.29.0

### Plugin version

6.1.1

### Node.js version

14.17

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13.0.1

### Description

I have defined my schema for the headers like below.

I have defined my schema for the headers like below.
```
const headers = {
  $id: "headers",
  type: "object",
  properties: {
    Authorization: { type: "string" },
    "x-country": {
      type: "string",
      transform: ["toUpperCase"],
      enum: ["IN", "US", "CH", "AR", "MX", "UY", "BR"]
    },
    "x-commerce": {
      type: "string",
      description: "Example values: 'FA', 'SO', 'TO' etc"
    }
  }
};
```

And added in one of my request header like
```
exports.createProductSizeSchema = {
  tags: ["Product Size"],
  summary: "This API is used to create  Product Sizes",
  description: "productSize should be Unique",
  headers: { $ref: "headers#" }
}
```

But while generating the definition, enum value is not coming up. PFB the screenshot
<img width="570" alt="Screenshot 2022-12-05 at 8 39 57 PM" src="https://user-images.githubusercontent.com/1974570/205672174-3a9eeaa3-0b8e-46fa-92f1-f07f1bd65e57.png">

Could you please let me know what I am doing wrong here?

### Steps to Reproduce

* Create a fastify instance and on registering plugins

```
const swagger = require("@fastify/swagger");

fastify.register(swagger, {
  routePrefix: "/documentation",
  swagger: {
    info: {
      title: "Test API",
      description: "API Docs",
      version: "0.1.0"
    },
    servers: [
    ],
    externalDocs: {
      url: "https://swagger.io",
      description: "Find more info here"
    },
    consumes: ["application/json"],
    produces: ["application/json"]
  },
  exposeRoute: true
});
```

### Expected Behavior

Expecting a response like
<img width="729" alt="Screenshot 2022-12-06 at 10 46 39 AM" src="https://user-images.githubusercontent.com/1974570/205822758-dca4cab6-1747-4188-97d8-799159271b30.png">

mcollina:
@climba03003 could you take a look?

climba03003:
The fix should be simple enough to allow more key for `header`.

[...]

PR #704, opened 2023-01-09
fix: Pass JSON Schema keywords into OpenAPI Header Spec

WoH:
#### Checklist

Instead of just fixing the problem for `enum` (see #701), I assume we can pass everything from the JSON Schema into the OpenAPI parameter's spec.
See the last OpenAPI test for an, admittedly, contrived scenario, which may however show why it may be short-sighted to only render the `string` specific keywords (`pattern`, `format`, `minLength`, `maxLength`). [This test is only for demo purposes]

Whenever the type is not `string`, it may make sense to make to explicitly set the `style` property.
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#parameter-object
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#style-values

In the end, this is similar to how we handle `query` parameters already, so this PR ended up basically just removing code.

While I'd always recommend separating the OpenAPI (Web) specific spec props (`style`, `explode`) from the JSON Schema, we pass it on in case of query params, so we should do that for headers the same way we treat query params anyway.

~~Currently, we do not pass `allowReserved` (OpenAPI v3), which I assume to be missing rather than an intended exclusion, in that case, should we add it?~~ Added

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -156,6 +156,7 @@ function plainJsonObjectToOpenapi3 (container, jsonSchema, externalSchemas, secu
   let toOpenapiProp
   switch (container) {
     case 'cookie':
+    case 'header':
     case 'query':
       toOpenapiProp = function (propertyName, jsonSchemaElement) {
         let result = {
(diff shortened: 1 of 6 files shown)


mcollina:
> Currently, we do not pass allowReserved (OpenAPI v3), which I assume to be missing rather than an intended exclusion, in that case, should we add it?

I think so!

WoH:
Added :)

mcollina (approved):
lgtm

Issue #737, opened 2023-07-02
Expose swagger/openapi version for transform hook


samchungy:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Add the openapi/swagger version to transform hook parameters. Unless there is a way to access the options passed into a plugin..?

### Motivation

I need to know what version of schema I am creating

### Example

```ts
transform: ({ schema, url, version }) => {
  if (version === '3.0.3') {
    // generate 3.0.3 schema
  }
}
```

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

Uzlopak:
we should maybe pass the whole options object to transform. version is too unspecific. In #732 it is requested to access the constraint, which also could contain version. Then people could get confused which version we actually mean. So better to pass config/options and constraints

Liam-Tait:
@mcollina I think this is a duplicate of https://github.com/fastify/fastify-swagger/issues/732 which I have made a pr for

samchungy:
> @mcollina I think this is a duplicate of #732 which I have made a pr for

Slightly different. I'm looking for the swagger/openapi version as opposed to the API version which is different

PR #738, opened 2023-07-03
feat: add api version filter based route version contraint

### LINKED ISSUE #732: Add ability to filter by version constraint
### Prerequisites

### 🚀 Feature Proposal

When generating a specification I want to be able to filter by the API version constraint so I can have multiple schema files with specific versions associated. For example, a schema for version 1 routes and version 2 routes.

There are two ways of adding this that come to mind:

**Add a specific filter for the `version` constraint to the Options object**

This approach makes it easy to integrate with find-my-way versioned routes using the same server logic. The options object would have an additional property such as `apiVersion` with a semver string that filters for only routes that match the server.

**Allow a route's `constraints` to be accessed within the `transform` function**

This would give users the option to use any constraint to to modify the schema, for version constraints this would be adding a step within the `transform` function to add the. Maximum flexibility but means it's user land code to ensure the versioning logic matches with find-my-way

### Motivation

When creating an API I would like to generate several API specifications, one for each major version.

### Example

```js
await fastify.register(require('@fastify/swagger'), {
  openapi: { ... },
  apiVersion: "1.0.0"
})
```
or
```js
await fastify.register(require('@fastify/swagger'), {
  swagger: { ... },
  transform: ({ schema, url, constraints }) => {
    if (!semver.satisfies(constraints.version, "1.0.0")) {
      schema.hide = true
    }
    return { schema, url }
  }
})
```

Liam-Tait:
Adds the ability to provide an `apiVersion` config value with a semver
range.

The route version is sourced from the [fastify version constraint](https://fastify.dev/docs/latest/Reference/Routes/#version-constraints).

Semver parsing is completed using the `semver` package `satisfies` function, which has been added as a
new dependency.

closes #732

#### Checklist

### DIFF
--- a/lib/util/common.js
+++ b/lib/util/common.js
@@ -4,6 +4,7 @@ const fs = require('fs')
 const path = require('path')
 const Ref = require('json-schema-resolver')
 const cloner = require('rfdc')({ proto: true, circles: false })
+const semverSatisfies = require('semver/functions/satisfies')
 const { rawRequired } = require('../symbols')
 const { xConsume } = require('../constants')
 
@@ -77,7 +78,7 @@ function addHook (fastify, pluginOptions) {
 }
 
 function shouldRouteHide (schema, opts) {
-  const { hiddenTag, hideUntagged } = opts
+  const { hiddenTag, hideUntagged, apiVersion, routeVersion } = opts
 
   if (schema && schema.hide) {
     return true
@@ -91,6 +92,10 @@ function shouldRouteHide (schema, opts) {
 
   if (tags.includes(hiddenTag)) {
     return schema.tags.includes(hiddenTag)
+  }
+
+  if (apiVersion && routeVersion && !semverSatisfies(routeVersion, apiVersion)) {
+    return true
   }
 
   return false
--- a/README.md
+++ b/README.md
@@ -209,17 +209,18 @@ An example of using `@fastify/swagger` with `static` mode enabled can be found [
 
 #### Options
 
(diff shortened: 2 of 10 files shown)


Uzlopak:
See discussion in #737 

Maybe pass the whole constraints object. Lets discuss this in #737

mcollina (approved):
lgtm

lib/util/common.js
```diff
@@ -93,6 +94,10 @@ function shouldRouteHide (schema, opts) {
     return schema.tags.includes(hiddenTag)
   }
 
+  if (apiVersion && routeVersion && !semverSatisfies(routeVersion, apiVersion)) {
+    return true
+  }
```
mcollina, line 99: By looking at this now, I think we could implemented a slightly more generic implementation that allows the user to provide a function instead, and we could add an example for that case.

Liam-Tait: Do you mean something like this?

```js
  filterRoutes: (route) => {
    // some custom logic
    return true;
  }
})
```

So for my use case it would be

```js
  filterRoutes: (route) => {
    if(route?.constraints?.version) {
      return semverSatisfies(route?.constraints?.version,'^1.0.0')
    }
    return true
  }
})
```

Liam-Tait:
If #739 goes through, this pr is not required as the functionality is possible using `transform`

mcollina:
Closing due to https://github.com/fastify/fastify-swagger/pull/739

PR #741, opened 2023-07-04
feat: add transformObject hook

### LINKED ISSUE #736: Transform hook for modifying full documentation
### Prerequisites

### 🚀 Feature Proposal

Add another `transform` (name TBC maybe `transformDocument`?) hook which exposes the generated documentation before it is published to allow for modification

### Motivation

I'm trying to create a plugin which brings together my library https://github.com/samchungy/zod-openapi for use with this. 

However, my library allows for dynamic registration of OpenAPI components via the schemas.

eg. you declare `z.string().openapi({ref: 'jobId')}` and it will automatically insert `{jobId: { type: 'string' }}` into the components and leave a $ref behind.

It also allows users to add Zod Objects directly to the components section and we will replace them with converted objects.

This presents a little bit of an issue because it doesn't seem like I can modify the components during the `transform` hook you currently expose, nor would I be able to allow for the same behaviour of inserting ZodObjects into the components.

So, if you would accept a PR for this, could I expose the full documentation object before it's served so third parties could modify documentation accordingly?

### Example

```ts
const transformDocument = (document) => {
  // do whatever you want with the document
  document.components = {};
  return document;
}
```

samchungy:
I've kept it similar to the `transform` hook to allow for additional parameters should we wish to add them in the future.

Resolves #736

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ await fastify.register(require('@fastify/swagger'), {
   }
 })
 ```
-
+<a name="register.options.transformObject"></a>
 #### Transform Object
 
 By passing a synchronous `transformObject` function you can modify the resulting `swaggerObject` or `openapiObject` before it is rendered.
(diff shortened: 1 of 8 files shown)


mcollina (approved):
lgtm

PR #751, opened 2023-08-23
fix(openapi): support nullish consts

WoH:
#### Checklist

### DIFF
--- a/test/spec/openapi/schema.js
+++ b/test/spec/openapi/schema.js
@@ -655,7 +655,11 @@ test('support "const" keyword', async t => {
           obj: {
             type: 'object',
             properties: {
-              constantProp: { const: 'my-const' }
+              constantProp: { const: 'my-const' },
+              constantPropZero: { const: 0 },
+              constantPropNull: { const: null },
+              constantPropFalse: { const: false },
+              constantPropEmptyString: { const: '' }
             }
           }
         }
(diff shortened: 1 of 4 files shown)


test/spec/openapi/schema.js
```diff
@@ -655,7 +655,9 @@ test('support "const" keyword', async t => {
           obj: {
             type: 'object',
             properties: {
-              constantProp: { const: 'my-const' }
+              constantProp: { const: 'my-const' },
```
Uzlopak, line 658: please also add a test for false and empty string.

Uzlopak (approved):
seems right, but I would like to have a second opinion by @ivan-tymoshenko

mcollina (approved):
lgtm

WoH:
> I'm not really deep into this library. The only thing I would check (add a test) is how it would work with `nullable` keyword and type `null`. For example:
> 
> ```json
> {
>   "type": "string",
>   "nullable": false,
>   "const": null
> }
> ```
> 
> and
> 
> ```json
> {
>   "type": "string",
>   "const": null
> }
> ```
> 
> Both these schemas are invalid (there is no value that would match them). You want to be sure that you handle cases like that correctly.

In both of these cases I'd argue it's not the job of this library to figure out what may have been implied, but just convert it to an equally useless (albeit OAS compatible) schema.

I would like to mention that one could make an argument for converting:
```json
 {
  "const": null
}
```

to:

```json5
 {
  "enum": [null],
  "nullable": true // with this added
}
```

but that is is not required from my understanding of the spec (which seems to be in line with the swagger-validator).

ivan-tymoshenko (approved):

Issue #749, opened 2023-08-19
Endpoint-specific schema transform


thorhj:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

fastify-swagger allows choose a [transform function](https://github.com/fastify/fastify-swagger#register.options.transform) in case you have a different schema type than the built-in one.

You can have different schema types for different endpoints, but to my knowledge fastify-swagger does not support this (I couldn't find anything for this at least). I would like to have an endpoint-specific transform function that differs from the "global" transform function. Something like this is what I imagine:

```js
fastify.get("/my-endpoint", {
  schema: { ... },
  config: {
    swaggerTransform: ( ... ) => { ... }
  }
},
async (request, reply) => { ... });
```

This `swaggerTransform` property should be able to indicate the "default" (no) transform or any other transform function.

### Motivation

In some of my projects I am [fastify-type-provider-zod](https://www.npmjs.com/package/fastify-type-provider-zod) which has its own transform function for this use-case.

However, I have found a need for having a specific endpoint NOT using zod schemas to define the schema. In my case, we would like to use a plugin that adds a new route-shorthand to fastify to easily create AWS SNS event webhooks correctly, and in this plugin the schema is implemented using the default schema declaration. When mixing schema types in the project, fastify-swagger will fail because it expects the routes to have zod schemas, but this one endpoint does not.

My current workaround is to hide the endpoint in question from swagger, but I would like to have it included, and to do that I think I need an endpoint-specific transform function for the endpoints in question.

Uzlopak:
Does swagger really expect zod schemas? 

@kibertoad

kibertoad:
@Uzlopak not swagger per se, but if type-provider-zod compiler and validator are registered, that will be the format of schemas that they will expect

kibertoad:
see https://github.com/turkerdev/fastify-type-provider-zod/blob/85675fad44c7679622d80fb1925f5901cc2f57f8/src/index.ts#L32

thorhj:
By the way, if you think this is a good idea and if we make a design for the interface (there are probably opinions on my suggestion), I could work on this feature myself.

MatanYadaev:
@mcollina Suggested this API: https://github.com/fastify/fastify-swagger/issues/727#issuecomment-1618258696

fredrikj31:
[...]

PR #752, opened 2023-08-25
adds endpoint-specific transforms

thorhj:
Addressing #749 

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,34 +1,39 @@
-import { FastifyPluginCallback, FastifySchema, RouteOptions, onRequestHookHandler, preHandlerHookHandler } from 'fastify';
-import { OpenAPI, OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from 'openapi-types';
+import {
+  FastifyPluginCallback,
+  FastifySchema,
+  RouteOptions,
+  onRequestHookHandler,
+  preHandlerHookHandler,
+} from "fastify";
+import { OpenAPI, OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from "openapi-types";
 
 /**
  * Swagger-UI Vendor Extensions
  * @see https://support.smartbear.com/swaggerhub/docs/apis/vendor-extensions.html#api-docs-x-tokenname
  */
-declare module 'openapi-types' {
+declare module "openapi-types" {
   namespace OpenAPIV3 {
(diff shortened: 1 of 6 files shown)


thorhj:
Two considerations for the reviewer:

1. The property for the transform function is called `swaggerTransform` but supports both OpenAPI and Swagger specs of course. Don't know if that is confusion and a better name should be added (or two separate properties).
2. The implementation allows the `swaggerTransform` to be given the value `null` which has its own specific meaning. Maybe this should be the string literal `"disable"` or something else to be more clear.

Uzlopak (changes requested):
missing typing tests.

thorhj:
> missing typing tests.

Added some tests in `types.test.ts`. Let me know if they should be written differently, haven't made such tests before (I'm usually working in Typescript directly 😉)

mcollina (approved):
lgtm

index.d.ts
```diff
@@ -52,8 +52,26 @@ declare module 'fastify' {
       [statusCode: string]: OpenAPIV3.ResponseObject['links'];
     }
   }
+  
+  interface FastifyContextConfig {
+    swaggerTransform?: SwaggerTransform | null;
+  }
 }
 
+type SwaggerTransform = <S extends FastifySchema = FastifySchema>({
```
Uzlopak, line 61: Why not
```typescript
type SwaggerTransform<S extends FastifySchema = FastifySchema> = ({
```

And is it still possible to specify the generic?

Uzlopak (changes requested):
I reviewed it now. 

swggerTransform null is quiet unintuitive. I would personally prefer false for deactivating.

index.d.ts
thorhj: I simply extracted the type from the global transform so I could re-use the same function type. I think it is correct; the function should be generic, not the type.

thorhj:
> I reviewed it now.
> 
> swggerTransform null is quiet unintuitive. I would personally prefer false for deactivating.

I agree, changed it to `false` instead of `null`.

index.d.ts
Uzlopak: Thanks for the clarification. Lets see if somebody complains... :))

fredrikj31:
When is this set to be released? 🤔

mcollina:
done now

PR #769, opened 2023-12-02
fix(openapi): hide auth header when set in securityScheme

### LINKED ISSUE #734: Authorization header is required while being set
### Prerequisites

### Fastify version

4.18.0

### Plugin version

8.6.0

### Node.js version

16

### Operating system

Linux

22.04

### Description

Is there a way to hide authorization header from OpenAPI 3.1.0 UI, so that I would not need to set a dummy unused value every time?

alt text

### Steps to Reproduce

```
import { fastify } from "fastify";
import fs from "@fastify/swagger";
import fsu from "@fastify/swagger-ui";

const app = fastify();

await app.register(fs, {
  openapi: {
    openapi: "3.1.0",
    components: {
      securitySchemes: {
        bearerAuth: {
          type: "http",
          scheme: "bearer",
        },
      },
    },
  },
});

await app.register(fsu);

app.get(
  "/",
  {
    schema: {
      headers: {
        type: "object",
        properties: { authorization: { type: "string" } },
        required: ["authorization"],
      },
      security: [{ bearerAuth: [] }],
    },
  },
  async (request, reply) => {
    return { hello: "world" };
  }
);

app
  .listen({ port: 3001 })
  .then(() => {
    console.log("listening");
  })
  .catch((err) => console.log(err));
```

### Expected Behavior

Authorization valued should not be required if already set or authorization property should be hidden in UI

aheckmann:
Treat the `http` `bearer` auth security scheme the same as other security schemes by removing the `authorization` header.

Fixes #734

#### Checklist

- [x] run `npm run test` and `npm run benchmark` <== `benchmark` doesn't exist so skipped that one.
- [x] documentation is changed or added <== N/A

### DIFF
--- a/examples/options.js
+++ b/examples/options.js
@@ -52,11 +52,16 @@ const openapiOption = {
           type: 'apiKey',
           name: 'apiKey',
           in: 'header'
+        },
+        bearerAuth: {
+          type: 'http',
+          scheme: 'bearer'
         }
       }
     },
     security: [{
-      apiKey: []
+      apiKey: [],
+      bearerAuth: []
     }],
     externalDocs: {
       description: 'Find more info here',
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -373,7 +373,10 @@ function prepareOpenapiMethod (schema, ref, openapiObject, url) {
   ]
     .reduce((acc, securitySchemeGroup) => {
       Object.keys(securitySchemeGroup).forEach((securitySchemeLabel) => {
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

Could the same issue show up with Swagger?

aheckmann:
@mcollina From what I could find, this lib already handles swagger properly b/c bearer auth does not have special handling in the spec - only OpenAPI 3 has the special bearer declaration where `in` is not specified.

mcollina (approved):
lgtm

Issue #535, opened 2022-01-29
"Examples" incorrectly rendered for OpenAPI 3.x spec 


maolimu:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.27.0

### Plugin version

4.13.0

### Node.js version

16.13.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.2 (21D49)

### Description

When adding an examples array to JSON Schema, when translating it to OpenAPI 3.x spec, the property seems to be misplaced: it is added under schemas but it should be next to it.

This causes a validation error in Swagger Editor, indicating the **examples** is not a valid property under **schemas** and it makes Swagger-UI not render the examples array.

### Steps to Reproduce

1) the schema is created with TypeBox
```typescript
export const PrototypesSchema = Type.Object(
    {
        foo: Type.String(),
        bar: Nullable(Type.String()),
    },
    {
        examples: [
            {
                foo: 'example 1',
                bar: null,
            },
            {
                foo: 'example 2',
                bar: 'some bar value',
            },
        ],
    }
);
```
2) the generated JSON-Schema is correct:
```json
{
   "examples": [
      {
         "foo": "example 1",
         "bar": null
      },
      {
         "foo": "example 2",
         "bar": "some bar value"
      }
   ],
   "type": "object",
   "properties": {
      "foo": {
         "type": "string"
      },
      "bar": {
         "type": "string",
         "nullable": true
      }
   },
   "required": [
      "foo",
      "bar"
   ]
}
```
3) The generated OpenAPI spec shows a validation error in Swagger Editor
```
Structural error at paths./experiments.put.requestBody.content.application/json.schema
should NOT have additional properties
additionalProperty: examples
Jump to line 29 
```

Indeed, the examples object is placed **below** schemas, but it should be **next** to it 
```yml
paths:
  /experiments:
    put:
      summary: experiments endpoint
      tags:
        - Misc
      requestBody:
        content:
          application/json:
            schema:
              examples:
                example1:
                  value:
                    foo: example 1
                    bar: null
                example2:
                  value:
                    foo: example 2
                    bar: some bar value
              type: object
              properties:
                foo:
                  type: string
                bar:
                  type: string
                  nullable: true
              required:
                - foo
                - bar
                - ```

### Expected Behavior

Place **examples** one level up, next to **schemas**, where it is correct

[...]

PR #541, opened 2022-02-03
feat(open-api): add required on body if necessary

Adibla:
#### Checklist

This MR add in open-api optional "required" property on schema body if exist at least one required properties in his schema.
It should solve https://github.com/fastify/fastify-swagger/issues/539 issue.
- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #772, opened 2023-12-13
fix: nested examples in OpenAPI schemas for request body & response

### LINKED ISSUE #771: Example values for individual request body fields are not compliant with OpenAPI specification
### Prerequisites

### Fastify version

4.24.3

### Plugin version

8.12.1

### Node.js version

21.3.0

### Operating system

macOS

Sonoma 14.1.1

### Description

Fastify requires `examples` field to be an array of values, which are subsequently transformed into valid OpenAPI format.

  > [`schemaToMedia`](https://github.com/fastify/fastify-swagger/blob/19a7513e538eae938a0c87890d424e68895f7cab/lib/spec/openapi/utils.js#L234C16-L234C16) is the function that performs this transformation
  > 1. If array contains one element, `examples` property is deleted and replaced with `example` containing a single value
  > 2. If array contains multiple elements, it's remapped to an object [according to the standard](https://swagger.io/specification/#example-object)

Everything works well in URL params, query params and other schemas that cannot be recursively nested, but when it comes to the request body, transformation is only performed [at the top level of the schema](https://github.com/fastify/fastify-swagger/blob/19a7513e538eae938a0c87890d424e68895f7cab/lib/spec/openapi/utils.js#L260). **Examples for each individual field are not transformed at all and stay in the original array form,** which is in fact not compliant with the OpenAPI standard.

The impact of this issue is that it breaks various related tooling. Two downstream issues that I encountered were:
 - Scalar swagger editor currently cannot interpolate examples in request body
 - `orval` fails schema validation

Example endpoint:

```typescript
fastify.post(
  "/",
  {
    schema: {
      summary: "Test",
      body: {
        type: "object",
        properties: {
          text: {
            examples: ["Lorem ipsum"],
            type: "string",
          },
        },
        required: ["text"],
      },
    },
  },
  handler
)
```

Output:

```json
{
  "openapi": "3.0.3",
  "info": {},
  "paths": {
    "/": {
      "post": {
        "summary": "Test",
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  "text": {
                    "examples": [
                      "Lorem ipsum"
                    ],
                    // ^^^ SHOULD NOT BE AN ARRAY
                    "type": "string"
                  }
                },
                "required": [
                  "text"
                ]
              }
            }
          },
          "required": true
        }
      }
    }
  },
  "servers": []
}
```

### Steps to Reproduce

1. Configure `fastify-swagger` in `openapi` mode
2. Add examples to individual fields of a request body like in the example above
3. Bingo (check the output json)

### Expected Behavior

Current implementation of `schemaToMedia` is sufficient, however it should be called recursively for each of the `properties` in the request body.

shfrmn:
### Summary

Fixes #771 

This PR addresses the following issue — example values in nested schemas of request body and response are not getting transformed into OpenAPI compliant format. For example, fastify schema below is not being processed correctly. Specifically `.examples` field is currently preserved as is on any nested properties, even though array is not a valid format for OpenAPI examples.
```javascript
{
  type: 'object',
  properties: {
    hello: {
      type: 'string',
      examples: ['world']
    }
  }
}
```

### OpenAPI spec for example values
I will spell out my current understanding of the standard in case it will help someone in the future.

- In general, example values can be in [two following formats](https://swagger.io/specification/#media-type-object)
  - `example: <value>`
  - `examples: {<value_name>: { value: <value>, ... }}` - [official documentation](https://swagger.io/specification/#example-object)
- However, `.examples` is not supported at the top level of `.schema` objects, in `.properties` of objects and `.items` of arrays. This means those are only allowed to have a single example.
  - [Official documentation](https://swagger.io/specification/#schema-object)
  - [Discussion in swagger parser issues](https://github.com/APIDevTools/swagger-parser/issues/93#issuecomment-412115097) - relevant because the parser actually doesn't accept `.examples` in cases I just outlined
  - [Previous discussion in this repo](https://github.com/fastify/fastify-swagger/issues/535#issuecomment-1079928296) - comment by @avdwio 
- Despite that, top-level of [request body](https://swagger.io/specification/#media-type-object) and [response](https://swagger.io/specification/#response-object) schemas can still have multiple examples, because it's possible to places `.examples` field outside of `schema` — this behavior is already implemented and covered by the test suite.
- Parameters (path, query and headers) can have either a single example or multiple examples (in the object format)

#### Code
- Introduced [`schemaTypeToNestedSchemas`](https://github.com/shfrmn/fastify-swagger/blob/7aebcb1046b5b0f454455b437a0ec99dc62b78ee/lib/spec/openapi/utils.js#L234) strategy that allows to extract all sub-schemas for different schema types.
- Introduced [`resolveSchemaExamples`](https://github.com/shfrmn/fastify-swagger/blob/7aebcb1046b5b0f454455b437a0ec99dc62b78ee/lib/spec/openapi/utils.js#L250) and [`resolveSchemaExamplesRecursive`](https://github.com/shfrmn/fastify-swagger/blob/7aebcb1046b5b0f454455b437a0ec99dc62b78ee/lib/spec/openapi/utils.js#L259C10-L259C40) functions that only take a single example (or `x-examples` unchanged).
- Introduced [`schemaToMediaRecursive`](https://github.com/shfrmn/fastify-swagger/blob/7aebcb1046b5b0f454455b437a0ec99dc62b78ee/lib/spec/openapi/utils.js#L288) function that works the same as `schemaToMedia`, but also recursively resolves examples inside the schema.
- Switched [`resolveBodyParams`](https://github.com/fastify/fastify-swagger/blob/19a7513e538eae938a0c87890d424e68895f7cab/lib/spec/openapi/utils.js#L254) and [`resolveResponse`](https://github.com/fastify/fastify-swagger/blob/19a7513e538eae938a0c87890d424e68895f7cab/lib/spec/openapi/utils.js#L290) to recursive implementation.

#### Tests
Overall, I found current test coverage for example values very lacking and introduced much more extensive set of tests. Because these tests are very repetitive I had to decide whether to generalize them or not. I opted for explicit repetitive testing of each combination: single/multiple example values, parameters/body/response, shallow/deeply nested schemas, strings/numbers/objects/arrays. I added some grouping to the tests and believe explicitly testing each use case is the most maintainable approach within the current testing setup, but imho using `fast-check` would definitely help with repetitiveness and test quality.

The combined test file diff on this PR is a bit of a spaghetti monster, so I broke it down into two commits that are much easier to review separately.

[Commit 1](https://github.com/shfrmn/fastify-swagger/commit/64947a2025e25c16c43a5607159acd7e77dbe23a) - removed some tests:
- Two tests [here](https://github.com/fastify/fastify-swagger/blob/19a7513e538eae938a0c87890d424e68895f7cab/test/spec/openapi/option.js#L446C32-L446C32) were against the specification (asserted incorrect output)
- Four other tests that I removed for clarity of the diff output and reintroduced in the next commit

[Commit 2](https://github.com/shfrmn/fastify-swagger/commit/0f4887903bd35e993cd9de45787fe9504614cf9d) - added 30 explicit tests for each significant combination of configuration

#### Documentation
- Updated example in the README to match the implementation

#### Checklist


test/spec/openapi/option.js
```diff
@@ -815,153 +1607,7 @@ test('move examples of parameters from component to media', async (t) => {
...
-  t.ok(schema)
-  t.ok(schema.properties.hello.examples)
-  t.same(schema.properties.hello.examples, [
-    ['a', 'b', 'c'],
-    ['d', 'f', 'g']
-  ])
-})
-
-test('uses examples if has property required in body', async (t) => {
+test('marks request body as required', async (t) => {
```
shfrmn, line 1610: This test had a wrong description, I fixed it since it mentions examples, but doesn't test anything related to them.

Originally introduced in https://github.com/fastify/fastify-swagger/pull/541

shfrmn:
Apologies for changes after I already requested a review, but I realized there was a better implementation.

My initial solution, that you can see in the previous commits, made changes to `schemaToMedia` function and introduced a boolean parameter that was used in recursion. The downsides of that approach are:
- The purpose of `schemaToMedia` is to augment schema into a media type object, but I was misusing it for transforming schema examples. In the process examples were first reassigned from schema to media and then back.
- Making changes to a function that's already in use is not ideal
- The second boolean parameter was awkward

The new solution, even though more LoC, has several advantages:
- Doesn't mess with `schemaToMedia`
- Explicit intent (transforms examples within schema objects)
- Extensible, because it allows to easily add support for more nested schema types in the future (e.g. `anyOf` and `allOf`). I'm leaving it out of the scope of this PR, but might submit another one later.
- Decouples extraction of nested schemas from example resolution

mcollina (approved):
lgtm

shfrmn:
@mcollina When would be a good time to merge this?

mcollina:
Now ;)

shfrmn:
@mcollina Sorry for bothering you, but 8.14 didn't get published after merging. Not sure if it's a CI issue or needs to be done manually.

climba03003:
@shfrmn Published.

Issue #773, opened 2023-12-14
Wildcard paths fails Swagger validation 


andreaspalsson:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.25.0

### Plugin version

8.12.1

### Node.js version

18.16.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

14.1.1

### Description

We are running into a issue when creating a schema and Swagger documentation for a wildcard route. 

When creating a route with a wildcard in the path, the swagger schema produced does not pass the validation when used on https://editor.swagger.io/. The problem is that URL does not match with the path parameters in the schema because the `*` has been renamed to `wildcard`.

We notice this when we started to validate our Swagger schema with https://editor.swagger.io/ and the following errors occurred
```
Semantic error at paths./id/{id}/star/{wildcard}
Declared path parameter "wildcard" needs to be defined as a path parameter at either the path or operation level
Jump to line 20

Semantic error at paths./id/{id}/star/{wildcard}.get.parameters.1.name
Path parameter "*" must have the corresponding {*} segment in the "/id/{id}/star/{wildcard}" path
Jump to line 31
```

We have also created a small repo that shows the issue https://github.com/Milkywire/fastify-swagger-wildcard

### Steps to Reproduce

Setup a route with a wildcard like this
```javascript
 fastify.get(
   "/id/:id/star/*",
   {
     schema: {
       params: {
         id: { type: "number" },
         "*": { type: "string" },
       },
     },
   },
   (request, reply) => {
     reply.send(request.params);
   }
 );
```

The documentation produced by @fastify/swagger will replace the `*` with `{wildcard}` in the path of the endpoint. So the documentation will look like this

```json
"/id/{id}/star/{wildcard}": {
  "get": {
    	"parameters": [
      	{
        	"schema": { "type": "number" },
        	"in": "path",
        	"name": "id",
        	"required": true
      	},
      	{
        	"schema": { "type": "string" },
        	"in": "path",
        	"name": "*",
        	"required": true
      	}
    	],
    	"responses": { "200": { "description": "Default Response" } }
  	}
}
```

Copy the JSON into  https://editor.swagger.io/ and you will get the following errors
```
Semantic error at paths./id/{id}/star/{wildcard}
Declared path parameter "wildcard" needs to be defined as a path parameter at either the path or operation level
Jump to line 20

[...]

PR #775, opened 2023-12-27
fix: * param renamed to wildcard

casantosmu:
This plugin transforms the route "/example/*" into "/example/{wildcard}". Therefore, the schema for a route with a wildcard should be defined as follows:

```javascript
fastify.get(
  "/example/*",
  {
    schema: {
      params: {
        wildcard: { type: "string" },
      },
    },
  },
  (request, reply) => {
    const path = request.params["*"];
    reply.send(path);
  }
);
```

This pull request allows defining the schema for a route with a wildcard as expected:

```javascript
fastify.get(
  "/example/*",
  {
    schema: {
      params: {
        "*": { type: "string" },
      },
    },
  },
  (request, reply) => {
    const path = request.params["*"];
    reply.send(path);
  }
);
```

This resolves the issue discussed in [this GitHub issue](https://github.com/fastify/fastify-swagger/issues/773).

#### Checklist

### DIFF
--- a/lib/util/format-param-url.js
+++ b/lib/util/format-param-url.js
@@ -65,9 +65,9 @@ function formatParamUrl (str) {
           state = 'regexp'
           level++
         } else if (char === '*') {
-          // * -> wildcard
+          // * -> {*}
           // should be exist once only
-          path += '{wildcard}'
+          path += '{*}'
         } else {
           path += char
         }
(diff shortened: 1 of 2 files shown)


casantosmu:
As mentioned in the [issue](https://github.com/fastify/fastify-swagger/issues/773), there was another potential solution: when transforming the Fastify schema to OpenAPI, updating the parameter "*" to "wildcard" could have been an option. Similar to the transformation applied to the URL. However, I believe it is simpler for everyone to maintain consistency by preserving the parameter names in the path rather than modifying them.

mcollina (approved):
lgtm

mcollina:
looking at the change, I don't think it's semver-major.

climba03003 (approved):
> looking at the change, I don't think it's semver-major.

Yes, changing the displayed pathname doesn't cause any problem.
It just not good for identify what it intended to do.

Note: The same thing would happen on `unnamed RegExp param`.

Eomm (approved):

PR #788, opened 2024-03-14
Updates for Fastify v5 

giovanniruzzi:
### What
* Update to fastify workflow v4.1.0
* Update dependencies to latest major versions
* Update unit tests

### Why
This PR aims to prepare fastify-swagger for the next version of fastify v5.

### Todo
- [ ] Validate changes on unit test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ on:
 
 jobs:
   test:
-    uses: fastify/workflows/.github/workflows/plugins-ci.yml@v3
+    uses: fastify/workflows/.github/workflows/plugins-ci.yml@v4.1.0
     with:
       license-check: true
       lint: true
--- a/package.json
+++ b/package.json
@@ -42,22 +42,22 @@
     "@apidevtools/swagger-parser": "^10.1.0",
     "@fastify/cookie": "^9.3.1",
     "@fastify/pre-commit": "^2.1.0",
-    "@types/node": "^20.11.27",
+    "@types/node": "^20.12.8",
     "fastify": "^4.26.2",
     "fluent-json-schema": "^4.2.1",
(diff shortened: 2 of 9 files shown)


giovanniruzzi:
I'd like to point out key changes in the unit tests. The upgrade to tap v18.5.6 means stricter `strictSame` checks. Often, schema properties now have an extra parameter `[Symbol('json-schema-resolver.ignore')]: true`, from [json-schema-resolver ref-resolver.js](https://github.com/Eomm/json-schema-resolver/blob/main/ref-resolver.js#L8). Tap v16 used [libtap](https://github.com/tapjs/libtap), but tap v18 uses `tcompare`, which is stricter about assertions. I had to use 't.match' instead of 't.strictSame'. I'm considering eliminating the Symbol somehow. Looking for guidance. 
Thanks

simoneb:
> I'd like to point out key changes in the unit tests. The upgrade to tap v18.5.6 means stricter `strictSame` checks. Often, schema properties now have an extra parameter `[Symbol('json-schema-resolver.ignore')]: true`, from [json-schema-resolver ref-resolver.js](https://github.com/Eomm/json-schema-resolver/blob/main/ref-resolver.js#L8). Tap v16 used [libtap](https://github.com/tapjs/libtap), but tap v18 uses `tcompare`, which is stricter about assertions. I had to use 't.match' instead of 't.strictSame'. I'm considering eliminating the Symbol somehow. Looking for guidance. Thanks

I think `t.match` is an acceptable alternative

gurgunday:
Hey, any updates?

You can use t.match or an alternative and we will review

giovanniruzzi:
@gurgunday updated
sorry for the long wait

gurgunday (approved):

gurgunday:
Thanks!

PR #814, opened 2024-08-01
corrected fastify autoload example

james-kaguru:
This pull request is to correct the fastify autoload example . While trying out I noted that calling
```js
await fastify.ready()
fastify.swagger()
```
was not needed

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,14 +136,19 @@ fastify.swagger()
 You need to register `@fastify/swagger` before registering routes.
 
 ```js
-const fastify = require('fastify')()
-const fastify = fastify()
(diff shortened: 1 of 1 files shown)


Uzlopak (changes requested):
Doesnt seem right

Blocking

Eomm:
I can foresee that the issue here is the top-level await usage
https://stackoverflow.com/questions/46515764/how-can-i-use-async-await-at-the-top-level

Use the `type: module` or write an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)

PR #817, opened 2024-08-23
fix: multiple alternative security objects in route schema

### LINKED ISSUE #816: Lists of Security Requirement Objects on an operation are not supported
### Prerequisites

### Fastify version

4.28.1

### Plugin version

8.15.0

### Node.js version

22.6

### Operating system

Linux

24.04

### Description

It seems that the security property definition inside FastifySchema doesn't cover the use case of multiple different security objects on a route schema ([spec. ref](https://swagger.io/specification/#security-requirement-object)):

```ts
interface FastifySchema {
    ...
    security?: ReadonlyArray<{ [securityLabel: string]: readonly string[] }>;
}
```

The relevant part would be:

> When a list of Security Requirement Objects is defined on the [OpenAPI Object](https://swagger.io/specification/#openapi-object) or [Operation Object](https://swagger.io/specification/#operation-object), only one of the Security Requirement Objects in the list needs to be satisfied to authorize the request.

It should then be possible to use both the following:

```ts
{
  // api_key AND bearer_token
  security: [{ api_key: [], bearer_token: [] }]
}
```

```ts
{
  // api_key OR bearer_token
  security: [{ api_key: [] }, { bearer_token: [] }]
}
```

but the "OR" solution doesn't work and throws the error:

```log
Types of property 'security' are incompatible.
    Type '({ bearer_token: never[]; api_key?: undefined; } | { api_key: never[]; bearer_token?: undefined; })[]' is not assignable to type 'readonly { [securityLabel: string]: readonly string[]; }[]'.
      Type '{ bearer_token: never[]; api_key?: undefined; } | { api_key: never[]; bearer_token?: undefined; }' is not assignable to type '{ [securityLabel: string]: readonly string[]; }'.
        Type '{ bearer_token: never[]; api_key?: undefined; }' is not assignable to type '{ [securityLabel: string]: readonly string[]; }'.
          Property 'api_key' is incompatible with index signature.
            Type 'undefined' is not assignable to type 'readonly string[]'.
```

### Link to code that reproduces the bug

_No response_

### Expected Behavior

It should be possible to use both solutions in the security property.

beryxz:
Fixes #816

Update the `security` property of `FastifySchema` to allow multiple security objects in a route schema.

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,7 +40,7 @@ declare module 'fastify' {
     consumes?: readonly string[];
     produces?: readonly string[];
     externalDocs?: OpenAPIV2.ExternalDocumentationObject | OpenAPIV3.ExternalDocumentationObject;
-    security?: ReadonlyArray<{ [securityLabel: string]: readonly string[] }>;
+    security?: ReadonlyArray<{ [securityLabel: string]: (readonly string[] | undefined) }>;
     /**
      * OpenAPI operation unique identifier
      */
@@ -52,7 +52,7 @@ declare module 'fastify' {
       [statusCode: string]: OpenAPIV3.ResponseObject['links'];
     }
   }
-  
+
   interface FastifyContextConfig {
     swaggerTransform?: SwaggerTransform | false;
   }
--- a/examples/options.js
+++ b/examples/options.js
@@ -281,6 +281,15 @@ const schemaSecurity = {
   }
 }
 
+const schemaSecurityMultipleAlterntives = {
+  schema: {
+    security: [
+      { apiKey: [] },
+      { bearerAuth: [] }
+    ]
+  }
+}
+
 const schemaConsumes = {
   schema: {
     consumes: ['application/x-www-form-urlencoded'],
@@ -376,6 +385,7 @@ module.exports = {
   schemaHeaders,
   schemaHeadersParams,
   schemaSecurity,
+  schemaSecurityMultipleAlterntives,
   schemaConsumes,
   schemaProduces,
   schemaCookies,
--- a/examples/options.js
+++ b/examples/options.js
@@ -23,6 +23,9 @@ const swaggerOption = {
         type: 'apiKey',
         name: 'apiKey',
         in: 'header'
+      },
+      basicAuth: {
+        type: 'basic'
       }
     },
     security: [{
@@ -281,11 +284,20 @@ const schemaSecurity = {
   }
 }
 
-const schemaSecurityMultipleAlterntives = {
+const schemaOpenapiSecurityMultipleAlterntives = {
   schema: {
     security: [
       { apiKey: [] },
       { bearerAuth: [] }
+    ]
+  }
+}
+
+const schemaSwaggerSecurityMultipleAlterntives = {
+  schema: {
+    security: [
+      { apiKey: [] },
+      { basicAuth: [] }
     ]
   }
 }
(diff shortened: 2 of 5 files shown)


Uzlopak:
please add some unit tests

beryxz:
Are these unit tests what you had in mind?

Uzlopak:
To be honest, I dont see any code change in the lib folder. So this is only about typings?

beryxz:
Yes, it's only about updating the type of the `security` property in the `FastifySchema` interface. The linked issue has some more detailed information.

jdagerydnnsoft:
When this fix will release?...

index.d.ts
```diff
@@ -40,7 +40,7 @@ declare module 'fastify' {
     consumes?: readonly string[];
     produces?: readonly string[];
     externalDocs?: OpenAPIV2.ExternalDocumentationObject | OpenAPIV3.ExternalDocumentationObject;
-    security?: ReadonlyArray<{ [securityLabel: string]: readonly string[] }>;
+    security?: ReadonlyArray<{ [securityLabel: string]: (readonly string[] | undefined) }>;
```
climba03003, line 43: It is weird to allows `undefined` here.

beryxz:
As @climba03003 rightly pointed out, I also think my changes are invalid, and this issue/PR should be closed.

I looked into my problem more thoroughly, and it was not actually caused by Fastify or Typebox but by a lack of explicit typing on my end.

To give a simplified context:
```ts
const app = fastify().withTypeProvider<TypeBoxTypeProvider>();

// FastifyRequestTypebox & FastifyReplyTypebox defined as in the official README

export const SomeOperationSchema = {
    summary: 'Do some operation',
    security: [{ bearer_token: [] }, { api_key: [] }],
    response: {
        204: NoContentResponse
    },
};

async function some_operation(
    request: FastifyRequestTypebox<typeof SomeOperationSchema>,
    reply: FastifyReplyTypebox<typeof SomeOperationSchema>,
) {
    return reply.status(204).send();
}

app.get('/', { schema: SomeOperationSchema }, some_operation);
```

This approach didn't work as the dynamic result of `typeof SomeOperationSchema` doesn't satisfy the `FastifySchema` type:
```ts
[{ bearer_token: [] }, { api_key: [] }]
```
gets parsed as
```ts
({ bearer_token: never[]; test?: undefined; } | { test: never[]; bearer_token?: undefined; })[]
```
thus, not matching the `security` property type defined in `FastifySchema`.

I fixed this by explicitly setting the type of the route schema:
```diff
- export const SomeOperationSchema = {
+ export const SomeOperationSchema: FastifySchema = {
```

---

Would this be an appropriate solution?

If so, could we improve the official [fastify/fastify-type-provider-typebox](https://github.com/fastify/fastify-type-provider-typebox?tab=readme-ov-file#type-definition-of-fastifyrequest-fastifyreply--typeprovider) README to include a note/warning or something that helps prevent this?

climba03003:
I am closing this PR because it is related to how TypeScript works and coding styles.

> What would be an appropriate solution?

1. Place it all together to allows TypeScript infer.

```ts
import Swagger from '@fastify/swagger'
import Fastify from 'fastify'

const fastify = Fastify()

await fastify.register(Swagger)

fastify.get('/', {
  schema: {
    security: [{ api_key: [] }, { bearer: [] }]
  }
}, () => { })
```

2. Make use of `as const` to bypass TypeScript auto reflection for array.

```ts
import Swagger from '@fastify/swagger'
import Fastify from 'fastify'

const fastify = Fastify()

await fastify.register(Swagger)

const schema = {
  security: [{ api_key: [] }, { bearer: [] }]
} as const

fastify.get('/', {
  schema
}, () => { })
```

Issue #795, opened 2024-03-26
Missing description when using $ref


melroy89:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.26.2

### Plugin version

8.14.0

### Node.js version

20.11.1

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Linux Mint 23.1

### Description

When adding shared schemas via `fastify.addSchema()` for response objects and using the id to access the schema like so: `422: { $ref: 'internalServerErrorResponse#' }`, result in a missing description.

Instead of the description displayed from the added schema, you will keep seeing "Default Response". When not using shared ref schemas, but adding the schema to the route schema, the description is taking into account. So it only doesn't work when using ref.

### Steps to Reproduce

1. Add shared schema for a response including a description, like this:

```js
import fastify from 'fastify'
import fastifySwagger from '@fastify/swagger'
import fastifySwaggerUi from '@fastify/swagger-ui'

const fastify = fastify()

fastify.addSchema({
  $id: 'validationError',
  description: 'Validation Error',
  type: 'object',
  properties: {
    statusCode: { type: 'number' },
    code: { type: 'string' },
    error: { type: 'string' },
    message: { type: 'string' }
  }
})

// Use OpenAPI v3 spec
await fastify.register(fastifySwagger, {
      mode: 'dynamic',
      openapi: { }
})
// Also add Swagger-UI plugin
await fastify.register(fastifySwaggerUi, {
  routePrefix: '/docs',
})
```

Using this schema by ID:

```js
fastify.delete('/:id', { schema: {
  description: 'Delete an object (example)',
  response: {
    422: {
      $ref: 'validationError#'
    },
}, async (req, reply) => {
  // ...
})
```

Going to the `/docs` for swagger, you will see:

![image](https://github.com/fastify/fastify-swagger/assets/628926/8c9d08f9-0455-43a1-8a5e-4e7500b44db7)

### Expected Behavior

The description of the schema should be put in the response description as well, instead of showing "Default Response" (which is the fallback description text).

### Code

I don't know the codebase, and `utils.js` is quite complex code. But I suspect to look around there: https://github.com/fastify/fastify-swagger/blob/fedf75c4f314fe9b37c81c88525f1ee1874335fa/lib/spec/openapi/utils.js#L342

melroy89:
[...]

PR #826, opened 2024-10-04
docs: Fallback to reference description for responses

Nelspike:
#### Checklist

This PR attempts to resolve the issue on https://github.com/fastify/fastify-swagger/issues/795 by resolving the `$ref` given in the schema and fetching that description in case there's none. PTAL! 🚀

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,6 +468,9 @@ fastify.get('/responseDescription', {
   }
 }, () => {})
 ```
+
+Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback.
+
 <a name="route.response.2xx"></a>
 ##### Status code 2xx
 Fastify supports both the `2xx` and `3xx` status codes, however Swagger (OpenAPI v2) itself does not.
--- a/README.md
+++ b/README.md
@@ -469,7 +469,7 @@ fastify.get('/responseDescription', {
 }, () => {})
 ```
 
-Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback. Note that at the moment, `$ref` will only be resolved by matching with `$id` and not through complex paths. Currently being addressed in [this Pull Request](https://github.com/fastify/fastify-swagger/pull/676).
+Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback. Note that at the moment, `$ref` will only be resolved by matching with `$id` and not through complex paths.
 
 <a name="route.response.2xx"></a>
 ##### Status code 2xx
--- a/README.md
+++ b/README.md
@@ -469,7 +469,7 @@ fastify.get('/responseDescription', {
 }, () => {})
 ```
 
-Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback.
+Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback. Note that at the moment, `$ref` will only be resolved by matching with `$id` and not through complex paths. Currently being addressed in [this Pull Request](https://github.com/fastify/fastify-swagger/pull/676).
 
 <a name="route.response.2xx"></a>
 ##### Status code 2xx
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -3,6 +3,7 @@
 const { readPackageJson } = require('../../util/read-package-json')
 const { formatParamUrl } = require('../../util/format-param-url')
 const { resolveLocalRef } = require('../../util/resolve-local-ref')
+const { resolveSchemaReference } = require('../../util/resolve-schema-reference')
 const { xResponseDescription, xConsume, xExamples } = require('../../constants')
 const { rawRequired } = require('../../symbols')
 const { generateParamsSchema } = require('../../util/generate-params-schema')
@@ -303,6 +304,11 @@ function resolveCommonParams (container, parameters, schema, ref, sharedSchemas,
   arr.forEach(swaggerSchema => parameters.push(swaggerSchema))
 }
 
+function findReferenceDescription (rawSchema, ref) {
+  const resolved = resolveSchemaReference(rawSchema, ref)
+  return resolved?.description
+}
+
 // https://swagger.io/docs/specification/describing-responses/
 function resolveResponse (fastifyResponseJson, produces, ref) {
   // if the user does not provided an out schema
@@ -327,7 +333,10 @@ function resolveResponse (fastifyResponseJson, produces, ref) {
     }
 
     const response = {
-      description: resolved[xResponseDescription] || rawJsonSchema.description || 'Default Response'
+      description: resolved[xResponseDescription] ||
+        rawJsonSchema.description ||
+        findReferenceDescription(rawJsonSchema, ref) ||
+        'Default Response'
     }
 
     // add headers when there are any.
--- a/lib/util/resolve-schema-reference.js
+++ b/lib/util/resolve-schema-reference.js
@@ -4,7 +4,7 @@ function resolveSchemaReference (rawSchema, ref) {
   const resolvedReference = ref.resolve(rawSchema, { externalSchemas: [ref.definitions().definitions] })
 
   // Ref has format `#/definitions/id`
-  const schemaId = resolvedReference?.$ref?.split('/')[2]
+  const schemaId = resolvedReference?.$ref?.split('/', 3)[2]
 
   if (schemaId === undefined) {
     return undefined
--- a/lib/util/resolve-schema-reference.js
+++ b/lib/util/resolve-schema-reference.js
@@ -1,0 +1,18 @@
+'use strict'
+
+function resolveSchemaReference (rawSchema, ref) {
+  const resolvedReference = ref.resolve(rawSchema, { externalSchemas: [ref.definitions().definitions] })
+
+  // Ref has format `#/definitions/id`
+  const schemaId = resolvedReference?.$ref?.split('/')[2]
+
+  if (schemaId === undefined) {
+    return undefined
+  }
+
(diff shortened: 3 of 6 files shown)


lib/spec/openapi/utils.js
```diff
@@ -326,8 +338,13 @@ function resolveResponse (fastifyResponseJson, produces, ref) {
       statusCode = statusCode.toUpperCase()
     }
 
+    const referenceJsonSchema = findReference(definitions, rawJsonSchema.$ref)
```
jean-michelet, line 341: I think this search should happen only if `rawJsonSchema.description` is falsy, not systematically.

lib/spec/openapi/utils.js
```diff
@@ -313,6 +324,7 @@ function resolveResponse (fastifyResponseJson, produces, ref) {
   const responsesContainer = {}
 
   const statusCodes = Object.keys(fastifyResponseJson)
+  const definitions = ref.definitions().definitions
```
jean-michelet, line 327: Can be done in the find function.

README.md
```diff
@@ -468,6 +468,9 @@ fastify.get('/responseDescription', {
   }
 }, () => {})
 ```
+
+Additionally, if you provide a `$ref` in your response schema but no description, if your `$ref` base schema includes a description, that will be used.
```
jean-michelet, line 472: I am not sure that's very clear. Do you think you should rephrase it?

jean-michelet (commented):
Thanks for the PR!

I don't know the code base so I can't approve myself.

jean-michelet:
Does it solve your bug @melroy89?

melroy89:
I directly let everything fall from my hands and looked into this PR! And using this utils.js file from the PR and override the file in `node_modules/@fastify/swagger/lib/spec/openapi/utils.js`, so I could directly test it on my existing code-base.

I'm using `addSchema()` method for shared schemas ([see docs](https://fastify.dev/docs/latest/Reference/Validation-and-Serialization/#adding-a-shared-schema)). Together indeed with `@fastify/swagger` with OpenAPI spec v3 and the `@fastify/swagger-ui` package for the swagger UI documentation you see below.

Too bad the issue with fallback description is still present 😢 :

image

Here such an example of a shared schema.

```js
export const notFoundErrorSchema = {
  $id: 'notFoundError',
  description: 'Not Found',
  type: 'object',
  properties: {
    message: { type: 'string' }
  },
  example: {
    message: 'Not found'
  }
}
```

Again, added via `fastify.addSchema(notFoundErrorSchema)` method.

I setup Fastify swagger like this:

```ts
 await fastify.register(fastifySwagger, {
      mode: 'dynamic',
      openapi: {
        openapi: '3.0.0',
        info: {
          title: 'Backend API Docs',
          description: 'RESTful backend API',
          version: '0.3.0',
        },
        servers: [
           ...
       ],
       tags: [
          ...
       ]
    }
})
```

Fastify Swagger UI is very simple to setup as well, basic is:

```ts
fastify.register(fastifySwaggerUi, {
      routePrefix: '/docs'
})
```

TLDR; issue is still present with fallback reference description in responses.

lib/spec/openapi/utils.js
```diff
@@ -303,6 +303,17 @@ function resolveCommonParams (container, parameters, schema, ref, sharedSchemas,
...
+// Keys to references could be different than the name of the $ref,
+// for example `def-0`
+function findReference (definitions, id) {
+  const schemaKey = Object.keys(definitions).find(key => {
+    const schema = definitions[key]
+    return schema.$id === id
+  })
+
+  return definitions[schemaKey] ?? {}
+}
```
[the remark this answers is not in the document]
Nelspike: Using `.values` makes sense actually, I'll change that 💪🏼 As for renaming the function, we could also go with that. My objective was having a function where you could resolve the reference for any other kinds of needs within this file for the future, but I can see how that's looking too far ahead and how it can be optimised.

Will make that change to return the description directly 👍🏼

lib/spec/openapi/utils.js
Nelspike: Since we don't need anything else, can definitely do so for now.

README.md
Nelspike: I can try, how's about something along the lines of:

```
Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback.
```

Nelspike:
First of all, @jean-michelet, thank you for the review! 🚀 

Secondly and equally thankful to @melroy89 for the detailed reproduction description 💪🏼 From what I can see from your code, the only reason why this wouldn't work is due to the trailing `#` on your `$ref`, which is `'notFoundError#'`. That `$ref` is absolutely correct, so we need to find a way to resolve these properly in this PR. If you want to give it a try without the trailing `#`, I think that would do the trick, at least for this example!

Nelspike:
@melroy89 - That's exactly what I meant in my reply, we need to consider those use cases because your reference is 100% correct 😅 The code I pushed in the PR didn't take that into account yet, but just pushed a change that does! I never meant that you should remove `#`, but that the code I had previously worked without it.

Nelspike:
Seems like the second use case you mentioned on your `EDIT` is blocked by this, @melroy89: https://github.com/fastify/fastify-swagger/pull/676

At the moment, seems like `definitions` is not supported within OpenAPI definitions on `fastify-swagger`, which difficults using refs like `commonResponse#/definitions/notFound`. _However_, your first use case with `#` at the end should be covered by this PR now 👍🏼 Once the other one's merged, it should be supported out of the box!

cc @asc11cat @Eomm @ivan-tymoshenko, tagging you folks for visibility since you were discussing on the other PR 😄

melroy89:
Yup works now! Thank you so much! I was looking for this for years.

@Nelspike I would like thank you by providing a donation! Any payment method will do.

lib/spec/openapi/utils.js
melroy89, line 1: What about the lib/spec/swagger/utils.js? Can this be added there as well easily? so OpenAPI folder is for OpenAI spec v3 basically. And the swagger folder is also 'OpenAPI' spec, but for older specifications like OpenAPI v2.

Nelspike: I can quickly take a look and see if we need it! 👍🏼

Nelspike:
@melroy89 - Thank you for your generosity and recognition! I'm totally okay doing this without a donation whatsoever, happy to help wherever I can 🎉

lib/spec/openapi/utils.js
Nelspike: And it's added on the latest commit 🎉

melroy89: Nice!

melroy89:
> @melroy89 - Thank you for your generosity and recognition! I'm totally okay doing this without a donation whatsoever, happy to help wherever I can 🎉

OK, if you change your mind. Let me know.

melroy89:
Now hopefully this PR gets merged somewhere before 2030 :)

lib/util/resolve-schema-reference.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+function resolveSchemaReference (rawSchema, ref) {
+  const resolvedReference = ref.resolve(rawSchema, { externalSchemas: [ref.definitions().definitions] })
+
+  // Ref has format `#/definitions/id`
+  const schemaId = resolvedReference?.$ref?.split('/')[2]
```
gurgunday, line 7: Nit:

```suggestion
  const schemaId = resolvedReference?.$ref?.split('/', 3)[2]
```

jean-michelet:
> Now hopefully this PR gets merged somewhere before 2030 :)

I don't think this type of comment is going to help us move any faster. If @climba03003 is ok with the code, I will merge it.

melroy89:
> > Now hopefully this PR gets merged somewhere before 2030 :)
> 
> I don't think this type of comment is going to help us move any faster. If @climba03003 is ok with the code, I will merge it.

No one can take a joke anymore.

gurgunday (approved):
lgtm

Haha don't worry @melroy89, I assure you we'll get this done this year :)

I just need another member (@climba03003 would be golden) to review

README.md
```diff
@@ -468,6 +468,9 @@ fastify.get('/responseDescription', {
   }
 }, () => {})
 ```
+
+Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback.
```
climba03003, line 472: Please state the `$ref` currently only support the format of matching `$id`.
It cannot lookup the fragment part.

climba03003 (changes requested):

lib/util/resolve-schema-reference.js
Nelspike: Oh yeah, nice catch! Thanks for the update 🎉

README.md
```diff
@@ -468,6 +468,9 @@ fastify.get('/responseDescription', {
   }
 }, () => {})
 ```
+
+Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback. Note that at the moment, `$ref` will only be resolved by matching with `$id` and not through complex paths. Currently being addressed in [this Pull Request](https://github.com/fastify/fastify-swagger/pull/676).
```
climba03003, line 472:
```suggestion
Additionally, if you provide a `$ref` in your response schema but no description, the reference's description will be used as a fallback. Note that at the moment, `$ref` will only be resolved by matching with `$id` and not through complex paths.
```

I bit later to say, but you don't need to point a PR in document. We don't knows which exact PR will resolve the issue in future. Other parts is good.

gurgunday (approved):
lgtm

I can release tonight, if anyone's free, feel free to merge and release

climba03003 (approved):

jean-michelet:
Release tonight plz @gurgunday!

PR #832, opened 2024-10-26
fix: soft fail fallback response description lookups

samchungy:
I tried fixing this issue in my library as raised in https://github.com/samchungy/fastify-zod-openapi/issues/172 

My library dynamically generates `$ref`s for users so they don't need to register anything upfront and plops them into the components section at the end.

As an example, my users will fill in the following:

```ts
responses: {
  200: z.string().openapi({ ref: 'Foo' })
}
```

The `transform` hook will output:

```ts
responses: {
  200: {
    $ref: '#/components/schemas/Foo'
  }
}
```

I tried dynamically injecting the components in the `transform` hook and even when I tried adding components into the fastifySwagger constructor this seems to not work.

```ts
    openapi: {
      components: {
		schema: {
		  foo: {
		    type: 'string'
		  }
		}
	}
});
```

This function does not seem to reference openapi components: https://github.com/fastify/fastify-swagger/blob/36a543cef71d32d434e07b610d6f0b233cbd566e/lib/util/resolve-schema-reference.js#L3-L18 and the rest of the code's support for components seem fairly limited.

Additionally this line in particular:
https://github.com/fastify/fastify-swagger/blob/36a543cef71d32d434e07b610d6f0b233cbd566e/lib/util/resolve-schema-reference.js#L7

only returns `schemas`.

I might revisit this repo to add proper openapi component support and start incrementally adding components to the `openapiObject` construct but for now I want to unblock my users.

#### Checklist

### DIFF
--- a/lib/util/resolve-schema-reference.js
+++ b/lib/util/resolve-schema-reference.js
@@ -10,7 +10,7 @@ function resolveSchemaReference (rawSchema, ref) {
     return undefined
   }
 
-  return resolvedReference.definitions[schemaId]
+  return resolvedReference.definitions?.[schemaId]
 }
 
 module.exports = {
--- a/test/spec/openapi/schema.js
+++ b/test/spec/openapi/schema.js
@@ -1022,6 +1022,63 @@ test('add default properties for url params when missing schema.params', async t
   })
 })
 
+test('support custom transforms which returns $ref in the response', async t => {
+  const customObject = {}
+  const opt = {
+    schema: {
+      response: {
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

samchungy:
Hey @gurgunday would you mind merging and releasing a patch version for this package?

I would love to cut a new release of my package, pinning the safe version for my consumers. Please and thank you! 🙏

mcollina (approved):
lgtm

PR #842, opened 2024-12-08
build(deps-dev): replace standard with neostandard

Fdawgs:
See https://github.com/fastify/fastify-cors/pull/340. This is a batch PR, so may have some issues. Please review changes.


Uzlopak:
fixed the linting issue, but also removed two unnecessary types which are in fastify-swagger-ui too.

PR #843, opened 2024-12-13
test: migrate from tap to node:test and c8

R-Campbell:
A few decisions while converting tests:
* `t.same` and `t.match` were converted to `t.assert.deepEqual`.
* `t.strictSame` was converted to `t.assert.deepStrictEqual`.
* `t.notOk` was converted to `t.assert.equal` comparing the coerced value (using `!!`) to `false`. We could have done `t.assert.ok(${value} === false)` but using `equal` with `false` felt more in the spirit of checking that for things that are not ok.
* `t.pass` was converted to `t.assert.ok(true, ${originalMessage})`
* Snapshots were removed and copied into the appropriate tests.

#### Checklist

- [x] run `npm run test`


.gitignore
```diff
@@ -148,8 +148,5 @@ yarn.lock
 .vscode
 .idea
 
-#tap files
-.tap/
-
```
dancastillo, line 153: Might be worth keeping this around for a bit just in case for people who ran tests

dancastillo (commented):
Can you update `t.assert.equal` to `t.assert.strictEqual` and `t.assert.deepEqual` to `t.assert.deepStrictEqual`

R-Campbell:
I made that update but ran into an issue with the Symbol that `json-schema-resolver` adds to the schema (specifically [json-schema-resolver.ignore](https://github.com/Eomm/json-schema-resolver/blob/main/ref-resolver.js)) and couldn't find a great way to reference that in the testing code. Instead I used `JSON.parse(JSON.stringify(...))` to strip the Symbols from the schema objects where needed (see [here](https://github.com/fastify/fastify-swagger/pull/843/files#diff-fc2b8af6daf81b33b3249e57b8ce40fcb66bdf3934c48e8aa67951c5dc47fed1R376) for an example). I don't know if that's the best solution, but it works.

I also found some places where `null` needed to be updated to `undefined` (see [here](https://github.com/fastify/fastify-swagger/pull/843/files#diff-33e69073fcabe579eb6d77ec4f8949b3b03931c5158f8b84c1269a6fe933bc28L452) for an example) as well as `t.fail` calls that I missed and changed to `t.assert.fail`.

test/mode/static.js
```diff
@@ -185,7 +203,7 @@ test('non-object specification.document throws an error', async (t) => {
   t.plan(1)
   const fastify = new Fastify()
 
-  t.rejects(async () => await fastify.register(fastifySwagger, config), new Error('specification.document is not an object'))
+  t.assert.rejects(async () => await fastify.register(fastifySwagger, config), new Error('specification.document is not an object'))
```
dancastillo, line 206:
```suggestion
  await t.assert.rejects(async () => await fastify.register(fastifySwagger, config), new Error('specification.document is not an object'))
```

test/spec/openapi/route.js
```diff
@@ -158,8 +158,8 @@ test('route options - produces', async (t) => {
 
   const api = await Swagger.validate(openapiObject)
   const definedPath = api.paths['/'].get
-  t.ok(definedPath)
-  t.match(definedPath.responses[200].content, {
+  t.assert.ok(definedPath)
+  t.assert.deepEqual(JSON.parse(JSON.stringify(definedPath.responses[200].content)), {
```
dancastillo, line 162: this works without `JSON.parse(JSON.stringify(...)`

test/spec/openapi/route.js
```diff
@@ -233,8 +233,8 @@ test('parses form parameters when all api consumes application/x-www-form-urlenc
 
   const api = await Swagger.validate(openapiObject)
   const definedPath = api.paths['/'].post
-  t.ok(definedPath)
-  t.match(definedPath.requestBody.content, {
+  t.assert.ok(definedPath)
+  t.assert.deepEqual(JSON.parse(JSON.stringify(definedPath.requestBody.content)), {
```
dancastillo, line 237: this works without `JSON.parse(JSON.stringify(...)`

dancastillo (commented):
Lets removed the `!!` to do a strict assert of `undefined` like this  `t.assert.strictEqual(!!value, undefined)`t.assert.strictEqual(value, undefined)` this will help explain behavior IMHO

dancastillo:
> I made that update but ran into an issue with the Symbol that `json-schema-resolver` adds to the schema (specifically [json-schema-resolver.ignore](https://github.com/Eomm/json-schema-resolver/blob/main/ref-resolver.js)) and couldn't find a great way to reference that in the testing code. Instead I used `JSON.parse(JSON.stringify(...))` to strip the Symbols from the schema objects where needed (see [here](https://github.com/fastify/fastify-swagger/pull/843/files#diff-fc2b8af6daf81b33b3249e57b8ce40fcb66bdf3934c48e8aa67951c5dc47fed1R376) for an example). I don't know if that's the best solution, but it works.
> 
> I also found some places where `null` needed to be updated to `undefined` (see [here](https://github.com/fastify/fastify-swagger/pull/843/files#diff-33e69073fcabe579eb6d77ec4f8949b3b03931c5158f8b84c1269a6fe933bc28L452) for an example) as well as `t.fail` calls that I missed and changed to `t.assert.fail`.

@Eomm wdyt here - `JSON.parse(JSON.stringify(...))` does remove the symbol from the object. Wondering if there is a preferred way to handle this?

dancastillo (approved):
LGTM

gurgunday (approved):
lgtm

Fdawgs (approved):

PR #852, opened 2025-01-08
perf(lib): replace accumulator spread with object.assign

Fdawgs:
Ended up stumbling across [this blog post over the holidays](https://www.richsnapp.com/article/2019/06-09-reduce-spread-anti-pattern), which covers why using spread (`{...obj}`) syntax in accumulators like `reduce` is a bad idea.

TL;DR: it causes a time complexity of `O(n^2)` instead of `O(n)` because it's creating an increasingly growing new object on every iteration. With object.assign, we're mutating the accumulator rather than creating a new object.

#### Checklist


gurgunday (approved):
Nice!

PR #872, opened 2025-05-15
feat(swagger): skip websocket routes

rozzilla:
Skip routes when `websocket: true`

### DIFF
--- a/test/spec/openapi/route.js
+++ b/test/spec/openapi/route.js
@@ -253,7 +253,6 @@ test('route options - websocket', async (t) => {
   const openapiObject = fastify.swagger()
 
   const api = await Swagger.validate(openapiObject)
-  console.log("api.paths['/']", api.paths['/'])
   t.assert.strictEqual(api.paths['/'], undefined, 'no schema created when websocket route')
 })
 
--- a/test/spec/openapi/route.js
+++ b/test/spec/openapi/route.js
@@ -229,17 +229,17 @@ test('route options - websocket', async (t) => {
 
   const opts = {
     schema: {
-      operationId: 'doSomething',
-      summary: 'Route summary',
+      operationId: 'websocket',
+      summary: 'Websocket route',
       tags: ['tag'],
-      description: 'Route description',
+      description: 'Route with websocket true',
       servers: [
         {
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
There are a lot of spurious changes - I think from prettier.

mcollina (approved):
lgtm

gurgunday (approved):
lgtm

test/spec/openapi/route.js
```diff
@@ -221,6 +221,42 @@ test('route options - extension', async (t) => {
...
+  }
+
+  fastify.get('/', opts, () => {})
+
+  await fastify.ready()
+
+  const openapiObject = fastify.swagger()
+
+  const api = await Swagger.validate(openapiObject)
+  console.log("api.paths['/']", api.paths['/'])
```
Uzlopak, line 256: Why console log?

Uzlopak:
I somehow wonder, why the websocket Attribute is so special. Are you setting it manually? Or is it because a Plugin is setting it?

Uzlopak:
Can we consider implementing it differently?

Maybe creating a Symbol. Like `Symbol.for("@fastify/swagger-hideRoute")`?

And in fastify websocket we set the route additionally to websocket true also that Symbol to true. And then we have a generic way to hide Routes. Without implementing for every possible plugin an exception. Also with that Symbol.for we loosely "bind" the plugins.

I would prefer this tbh..

Uzlopak:
@mcollina 

What do you think about my proposal?

mcollina:
We already have the known `hide: true` property as documented in the README. If you prefer that could be automatically added in https://github.com/fastify/fastify-websocket/blob/d6348f0c04a73fb27c932d842168a894ea075cca/index.js#L155.

rozzilla:
> We already have the known `hide: true` property as documented in the README. If you prefer that could be automatically added in https://github.com/fastify/fastify-websocket/blob/d6348f0c04a73fb27c932d842168a894ea075cca/index.js#L155.

@mcollina @Uzlopak [here is the PR](https://github.com/fastify/fastify-websocket/pull/336)

PR #873, opened 2025-05-16
chore(ci): fix node 24 tests

rozzilla:
As by title


Uzlopak:
so the test files are now extracted?

rozzilla:
> so the test files are now extracted?

@Uzlopak only the typescript ones. The issue was that when running with Node 24, typescript tests were picked up by the test runner, and therefore CI/CD was failing since such files should run with `tsd`, not `node --test`

rozzilla:
@Uzlopak [see here](https://github.com/fastify/fastify-swagger/actions/runs/15068876349/job/42359941308?pr=872#step:5:360)

mcollina (approved):
lgtm

This is unfortunately needed to run everything in node v24 - it will try to run those ts files as normal tests otherwise.

mcollina:
cc @Uzlopak

mcollina:
ah, that can work as well. @rozzilla would you mind?

rozzilla:
@mcollina @Uzlopak done

mcollina (approved):
lgtm

Uzlopak (approved):
Thank you :)

LGTM

Issue #894, opened 2025-09-04
requestBody.required not set to true when route has body schema with only optional properties


khokm:
### Fastify version

5.5.0

### Plugin version

9.5.1

### Node.js version

22.14.0

### Operating system

Windows 10

### Description

When a route has a `schema.body` defined, Fastify requires that the request includes a body (at least `{}`), even if all properties inside that body are optional.

However, the OpenAPI schema generated by `@fastify/swagger` does not correctly reflect this requirement. Currently, `requestBody.required` is set to `true` only when there are required properties inside the object. If the object has only optional properties, then `requestBody.required` is omitted.

Example:
```ts
import { fastifySwagger } from "@fastify/swagger";
import { JsonSchemaToTsProvider } from "@fastify/type-provider-json-schema-to-ts";
import { fastify } from "fastify";

const server = fastify().withTypeProvider<JsonSchemaToTsProvider>();

(async function () {
  await server.register(fastifySwagger, {
    openapi: {
      openapi: "3.1.0",
    },
  });

  server.put(
    "/",
    {
      schema: {
        body: {
          type: "object",
          properties: {
            hello: { type: "string" },
          },
        },
      },
    },
    async (req) => {
      console.log(req.body.hello?.length); // hello is optional but the body is not.
    },
  );

  await server.ready();

  const swagger = server.swagger();
})();
```

Expected OpenAPI output:
```json
"requestBody": {
  "required": true,
  "content": {
    "application/json": {
      "schema": {
        "type": "object",
        "properties": {
          "hello": { "type": "string" }
        }
      }
    }
  }
}
```

Actual OpenAPI output:
```json
"requestBody": {
  "content": {
    "application/json": {
      "schema": {
        "type": "object",
        "properties": {
          "hello": { "type": "string" }
        }
      }
    }
  }
}
```

In Fastify, the body itself is required if schema.body is defined.

The generated OpenAPI spec incorrectly marks it as optional, which can mislead client code generators and API consumers.

### Expected Behavior

If a route has `schema.body`, `requestBody.required` should always be set to true, regardless of whether any properties inside the `body` are required.

### Current fix using `transformObject`
```ts
transformObject: (documentObject) => {
    assert("openapiObject" in documentObject);

    const { openapiObject } = documentObject;

    for (const pathKey in openapiObject.paths) {
        const pathItem = openapiObject.paths[pathKey];
        if (!pathItem) continue;

[...]

PR #903, opened 2025-11-22
fix(openapi): always set requestBody.required to true when schema.body exists

bas0N:
Fastify requires a body when schema.body is defined, even if all properties are optional. This ensures the OpenAPI spec correctly reflects Fastify's runtime behavior. This solves the issue #894

## Problem

When a route has `schema.body` defined, Fastify requires that the request includes a body (at least `{}`), even if all properties inside that body are optional. However, the OpenAPI schema generated by `@fastify/swagger` was only setting `requestBody.required` to `true` when there were required properties inside the object. If the object had only optional properties, then `requestBody.required` was omitted, incorrectly marking the request body as optional.

This mismatch between the OpenAPI spec and Fastify's actual runtime behavior can mislead client code generators and API consumers.

## Solution

This fix ensures that `requestBody.required` is always set to `true` when `schema.body` is defined, regardless of whether any properties inside the body are required. The fix sets `required: true` when creating the `requestBody` object in `prepareOpenapiMethod`, making it explicit that the body is required whenever a body schema is defined.

## Example

**Before (incorrect):**

```json
{
  "requestBody": {
    "content": {
      "application/json": {
        "schema": {
          "type": "object",
          "properties": {
            "hello": { "type": "string" }
          }
        }
      }
    }
  }
}
```

**After (correct):**

```json
{
  "requestBody": {
    "required": true,
    "content": {
      "application/json": {
        "schema": {
          "type": "object",
          "properties": {
            "hello": { "type": "string" }
          }
        }
      }
    }
  }
}
```

## Changes

- Set `required: true` when creating `requestBody` object in `prepareOpenapiMethod`
- Added test case for body with only optional properties
- Updated existing tests to expect `required: true` when `schema.body` is defined

Fixes the issue where the OpenAPI spec incorrectly showed `requestBody` as optional when Fastify actually requires it at runtime.

#### Checklist

- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -281,10 +281,6 @@ function resolveBodyParams (opts, body, schema, consumes, ref) {
       body.content[consume] = media
     })
 
-    if (resolved?.required?.length) {
-      body.required = true
-    }
-
     if (resolved?.description) {
       body.description = resolved.description
     }
@@ -475,7 +471,9 @@ function prepareOpenapiMethod (opts, schema, ref, openapiObject, url) {
     if (schema.externalDocs) openapiMethod.externalDocs = schema.externalDocs
     if (schema.querystring) resolveCommonParams(opts, 'query', parameters, schema.querystring, ref, openapiObject.definitions, securityIgnores.query)
     if (schema.body) {
-      openapiMethod.requestBody = { content: {} }
+      // Fastify requires a body (at least {}) when schema.body is defined,
+      // so requestBody.required should always be true
+      openapiMethod.requestBody = { required: true, content: {} }
       resolveBodyParams(opts, openapiMethod.requestBody, schema.body, schema.consumes, ref)
     }
     if (schema.params) resolveCommonParams(opts, 'path', parameters, schema.params, ref, openapiObject.definitions)
--- a/test/spec/openapi/option.test.js
+++ b/test/spec/openapi/option.test.js
@@ -1660,6 +1660,38 @@ test('marks request body as required', async (t) => {
   t.assert.deepStrictEqual(requestBody.required, true)
 })
 
+test('marks request body as required even when all properties are optional', async (t) => {
+  t.plan(3)
+  const fastify = Fastify()
+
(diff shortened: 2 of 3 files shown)


bas0N:
@Fdawgs when you get a moment, could you please take a look at this PR? Thanks!

Fdawgs (approved):

ilteoood (commented):
LGTM
but honestly I think that on the user perspective this could be considered as a breaking change.
Imagine if someone is generating TS types from the fastify's schema: objects that before were not required, now are flagged as required 🤔

gurgunday (approved):
lgtm

bozzaj:
Hello team! Any chance that this can be merged sometime soon? Regarding the notes above regarding this being a breaking change. It's important to note that this package is specific to Fastify, and the current implementation is actually broken. For quite some time we've been manually fixing things. 

@ilteoood mentioned that "objects that before were not required are now flagged as required". While that's correct when focusing only on this package, it's important to state that the objects that were not required were actually ALWAYS required due to how Fastify works. In any case where there was a body defined, the Fastify API would generate a 400 error if no body was sent. So while the generated code that came out of the OpenAPI spec would have an optional body, in reality it has never been optional. The generated API from this plugin has always been broken in this respect, so any usage of the API that backed the generation would always fail if there was a body defined.

While it could be said that following proper semver processes, this could be considered a breaking change, I would challenge that simply because the current generated code is broken - it does not create a valid specification based on how Fastify works. Since this is a plugin specific to Fastify, and everyone using Fastify MUST pass at least an empty object when a body is defined, this really is fixing a bug so the generated code matches the expectations of Fastify.

Either way - bug fix or major version - it would be nice to not have to patch the package to provide valid functionality, so I appreciate if this PR could be looked at again!

Sorry for the long-winded comment!

gurgunday (approved):
lgtm

can someone else from @fastify/plugins review it too?

ilteoood (approved):

PR #910, opened 2026-02-05
fix(openapi): keep required query flag for  params

### LINKED ISSUE #902: Query Parameters Are Not Marked As `required: true` When Using `Ref`
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.6.2

### Plugin version

9.6.0

### Node.js version

24.11.0

### Operating system

Windows

10

### Description

When using a schema reference with fastify query model definition the resulting openapi.json does not specify that the property is required.

```js
// 1. define some sort of reusable schema
fastify.addSchema({
  $id: "CoringUploadTypeApiModel",
  type: "string",
  enum: ["health_safety", "coring"],
});

fastify.get(
  "/some-route",
  {
    schema: {
    query: {
        type: "object",
        // 2. specify thing property is required
        required: ["thing"],
        properties: {
        // 3. define thing property with $ref to reusable schema
        thing: {
            anyOf: [
            { type: "string", const: "health_safety" },
            { type: "string", const: "coring" },
            ],
            $ref: "CoringUploadTypeApiModel",
        },
        other: { type: "string" },
        },
    },
    response: {
        200: {
        type: "object",
        properties: {
            hello: { type: "string" },
        },
        },
    },
    },
  },
  (req, reply) => {
    reply.code(200).send({ hello: "world" });
  });
```

the resulting model definition when calling `.swagger()` is:

```json
{
  "openapi": "3.0.0",
  "info": {
    "version": "9.6.1",
    "title": "@fastify/swagger"
  },
  "components": {
    "schemas": {
      "def-0": {
        "type": "string",
        "enum": [
          "health_safety",
          "coring"
        ],
        "title": "CoringUploadTypeApiModel"
      }
    }
  },
  "paths": {
    "/some-route": {
      "get": {
        "parameters": [
          {
            "schema": {
              "$ref": "#/components/schemas/def-0"
            },
            "in": "query",
            "name": "thing"
          },
          {
            "schema": {
              "type": "string"
            },
            "in": "query",
            "name": "other",
            "required": false
          }
        ],
        "responses": {
          "200": {
            "description": "Default Response",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "hello": {
                      "type": "string"
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

notice how `thing` does not have `required: true`

### Link to code that reproduces the bug

https://github.com/cmcnicholas/fastify-optional-query-param-ref/blob/main/index.js

### Expected Behavior

_No response_

mcollina:
## Summary
- preserve `required` for query/header/cookie params even when the property schema uses `$ref`
- add regression test for issue #902 (`renders required query parameter when property is a $ref`)

## Root cause
OpenAPI parameter conversion was reading `required` from the converted property schema. For `$ref` properties, conversion strips sibling keys, so `required` could be lost.

## Fix
Read `required` from the pre-conversion local schema map and pass it explicitly when building parameter objects.

Fixes #902

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,3 +153,9 @@ yarn.lock
 
 # static files
 static
+
+# AI assistant local config
+.pi
+.claude
+AGENTS.md
+CLAUDE.md
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -134,17 +134,18 @@ function convertExamplesArrayToObject (examples) {
 // For supported keys read:
 // https://swagger.io/docs/specification/describing-parameters/
 function plainJsonObjectToOpenapi3 (opts, container, jsonSchema, externalSchemas, securityIgnores = []) {
-  const obj = convertJsonSchemaToOpenapi3(opts, resolveLocalRef(jsonSchema, externalSchemas))
+  const localJsonSchema = resolveLocalRef(jsonSchema, externalSchemas)
+  const obj = convertJsonSchemaToOpenapi3(opts, localJsonSchema)
   let toOpenapiProp
   switch (container) {
     case 'cookie':
     case 'header':
     case 'query':
-      toOpenapiProp = function (propertyName, jsonSchemaElement) {
+      toOpenapiProp = function (propertyName, jsonSchemaElement, required) {
         let result = {
           in: container,
           name: propertyName,
-          required: jsonSchemaElement.required
+          required
         }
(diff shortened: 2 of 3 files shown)


gurgunday (approved):
lgtm

PR #921, opened 2026-04-18
fix: hoist nested definitions to top-level components/schemas

### LINKED ISSUE #639: JSON Schema Definitions not work

Upadhyay-Om:
Fixes #639

## Problem

When using `addSchema()` with nested `definitions`, fastify-swagger silently deletes the `definitions` block during OpenAPI conversion. Any `$ref` pointing into those definitions ends up dangling — Swagger UI throws a "could not resolve pointer" error even though Fastify's runtime handles it perfectly fine.

## Fix

Two things needed fixing:

- Before deleting `definitions`, each nested entry now gets hoisted up to top-level `components/schemas` under the key `{parentName}-{defName}`. The deletion then becomes safe since nothing is lost.
- The `$ref` rewrite regex was too strict — it couldn't handle schema IDs that contain slashes (like URLs). Switched the capture group to a greedy match so it handles the full ID correctly.

## Test

Added a test in `test/spec/openapi/refs.test.js` using the exact reproduction case from the issue — registers a schema with nested `definitions`, references it via `$ref` in a route, and asserts both that the hoisted schema appears in `components/schemas` and that the `$ref` path is rewritten correctly. All existing tests still pass at 100% coverage.

Related: #676

#### Checklist
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11) and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -535,7 +535,13 @@ function convertJsonSchemaToOpenapi3 (opts, jsonSchema) {
     }
 
     if (key === '$ref') {
-      openapiSchema.$ref = value.replace('definitions', 'components/schemas')
+      // nested definitions ref pattern: #/definitions/def-0/definitions/foo
+      const nestedMatch = value.match(/^#\/definitions\/(.+)\/definitions\/([^/]+)$/)
+      if (nestedMatch) {
+        openapiSchema.$ref = `#/components/schemas/${nestedMatch[1]}-${nestedMatch[2]}`
+      } else {
+        openapiSchema.$ref = value.replace('#/definitions/', '#/components/schemas/')
+      }
       continue
     }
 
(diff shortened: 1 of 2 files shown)


mikaelkaron:
Not had a close look at the code yet but wanted to make you aware of #676. I don't see you having the same problems but could be worth while taking a quick read of the comments.

PR #932, opened 2026-07-13
fix: reduce the numbers of onReady hooks

climba03003:
Partial Address https://github.com/fastify/fastify/issues/6843

- Reduce the numbers of `onReady` hook from n to 1.

#### Checklist

- [ ] run `npm run test && npm run benchmark --if-present`
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)


lib/util/add-hook.js
```diff
@@ -37,27 +38,27 @@ function addHook (fastify, pluginOptions) {
...
-        sharedSchemasMap.set(schemaId, allSchemas[schemaId])
-      }
-      done()
-    })
+    instances.push(instance)
   })
 
   fastify.addHook('onReady', (done) => {
     hookRun = true
+    for (const instance of instances) {
```
climba03003, line 46: The most outer `onReady` run the last.
So, we can use a single `onReady` hook to register all schemas including the sub-instances.

climba03003:
The schemas registration is tested on

- https://github.com/fastify/fastify-swagger/blob/main/test/spec/swagger/refs.test.js
- https://github.com/fastify/fastify-swagger/blob/main/test/spec/openapi/refs.test.js

So, it doesn't require a new test.

mcollina (approved):
lgtm

Issue #770, opened 2023-12-03
Custom "examples" names and summaries for request body, params, etc


mouhannad-sh:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Support distinct examples name for request body, params, etc.. 

### Motivation

The openapi v3 spec supports adding a distinct name and summary for each example as shown here 
https://swagger.io/docs/specification/adding-examples/

```yaml
parameters:
  - in: query
    name: limit
    schema:
      type: integer
      maximum: 50
    examples:       # Multiple examples
      zero:         # Distinct name
        value: 0    # Example value
        summary: A sample limit value  # Optional description
      max: # Distinct name
        value: 50   # Example value
        summary: A sample limit value  # Optional description
```

<img width="446" alt="image" src="https://github.com/fastify/fastify-swagger/assets/18495740/da44afa8-c6bd-470b-a4ae-fc1bcbbf332c">

unfortunately, it seems like an object format which is not allowed in the current implementation due to the dependency on `fast-json-stringify` and it's schema validator
 
https://github.com/fastify/fast-json-stringify/blob/4f6f7b809fc7148a40e8f07cc81a658757e81b68/lib/schema-validator.js#L203C1-L206C2

There's probably a good reason for that. but I'm wondering if there's a way around it? 

I'm happy to contribute once I have a decent understanding on what can be done

### Example

The current convertor implementation in this library forces examples to be named as `example 1`, `example 2`, ... 

https://github.com/fastify/fastify-swagger/blob/5277980820f96a7831d026dbc280f2d3a2972f4f/lib/spec/openapi/utils.js#L148-L158

Eomm:
Thanks for reporting!
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

jiroler:
I faced to the same problem and found this solution: https://github.com/fastify/fastify-swagger#add-examples-to-the-schema

mouhannad-sh:
thanks @jiroler I'll test it out and report back. if that doesn't work then I might submit a PR !

mouhannad-sh:
I can confirm that it works !
in my case I'm using it with [Typebox](https://github.com/sinclairzx81/typebox)

```ts
fastify({
    ajv: {
      plugins: [
        function (ajv) {
          ajv.addKeyword({ keyword: 'x-examples' })
        }
      ]
    }
  })
  
[...]

PR #616, opened 2022-06-11
Allow custom schema examples in OpenAPI format

mxck:
Hi! 
I added the ability to pass examples directly through the "x-examples" field (check this https://github.com/fastify/fastify-swagger/issues/535#issuecomment-1079928296)  and added documentation about examples in general.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #935, opened 2026-07-29
fix/examples: Fix OpenAPI "examples"

gevalo1:
> Note: AI-assisted (Claude Opus 5) while writing, human-reviewed and human-tested.

## What this fixes

`examples` arrays are always rewritten into the OpenAPI 3.0 shape, even when the
document declares `openapi: 3.1.0`.

In 3.0 that rewrite is correct and necessary: the Schema Object there is a subset
of JSON Schema draft 4 and only has a singular `example`, while `examples` is a
map of Example Objects that lives on the Media Type or Parameter Object. So a
single example is moved into `example`, and multiple examples become a named map
one level up.

In 3.1 the Schema Object *is* JSON Schema 2020-12. `examples` is a plain array,
valid at any depth, and `example` is deprecated. Running the 3.0 rewrite against a
3.1 document therefore is not correct:

- a deprecated keyword is emitted where the modern one was already valid
- values are silently dropped, only `examples[0]` survives
- the output is inconsistent, because schemas the recursion does not reach (for
  example anything under `anyOf`/`oneOf`) keep their arrays, so one document ends
  up mixing both styles

This caused issues for us: we generate a 3.1 document for our REST docs and
were using a `pnpm patch` for this plugin to get a valid OpenAPI schema.

## What I changed

The example handling now takes the target version into account. For `3.1.x` the
arrays are left exactly as written, at every depth, and nothing is hoisted. Every
other version keeps the current behaviour, unchanged and untouched.

There is no new plugin option. The version already selects this, which matches what was asked for in #625:

> We should add support to OAS v3.1 too but keep the v3.0 line as-is

> We differentiate with an option to the plugin, check with how swagger vs oas are handled

`x-examples` is deliberately untouched and keeps producing the named Example Object
map in both versions, so the use case from #616 and #770 still works. That remains
the way to attach names, summaries and descriptions to examples.

## The first commit is separate

While testing we found that parameter schemas were converted with `schemaToMedia`
instead of `schemaToMediaRecursive`, so nested `examples` arrays inside a parameter
were never resolved. An array query parameter like this:

```js
querystring: {
  type: 'object',
  properties: {
    tags: { type: 'array', items: { type: 'string', examples: ['a', 'b'] } }
  }
}
```

#### Checklist

- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)


gevalo1:
Just for reference, this is what our `pnpm patch` looks like currently, so that we get a valid OpenAPI 3.1 schema:

```
diff --git a/lib/spec/openapi/utils.js b/lib/spec/openapi/utils.js
index 164cf1b05edfcbeeca352ac8fbb3f13fb09a6902..1ab5bed2c9c04a507a13740d88402fefd402f799 100644
--- a/lib/spec/openapi/utils.js
+++ b/lib/spec/openapi/utils.js
@@ -223,12 +223,12 @@ const schemaTypeToNestedSchemas = {
 }
 
 function resolveSchemaExamples (schema) {
-  const example = schema[xExamples] ?? schema.examples?.[0]
-  if (typeof example !== 'undefined') {
-    schema.example = example
-  }
-  delete schema[xExamples]
-  delete schema.examples
+  // const example = schema[xExamples] ?? schema.examples?.[0]
+  // if (typeof example !== 'undefined') {
+  //   schema.example = example
+  // }
+  // delete schema[xExamples]
+  // delete schema.examples
 }
 
 function resolveSchemaExamplesRecursive (schema) {
@@ -244,8 +244,8 @@ function schemaToMedia (schema) {
   const media = { schema }
 
   if (schema.examples?.length === 1) {
-    media.example = schema.examples[0]
-    delete schema.examples
+    // media.example = schema.examples[0]
+    // delete schema.examples
   } else if (schema.examples?.length > 1) {
     media.examples = convertExamplesArrayToObject(schema.examples)
     // examples is invalid property of media object schema
```

gevalo1:
Hi @mcollina, @Fdawgs, @climba03003, @Tony133
(Sorry for the ping)

Would appreciate a review when someone has time, happy to discuss and/or make any requested changes!

README.md
```diff
@@ -1083,7 +1085,49 @@ Generates this in the OpenAPI v3 schema's `paths`:
...
 
-Use the `x-examples` field to set names or add descriptions to schema examples in [OpenAPI format](https://swagger.io/specification/#example-object):
+In OpenAPI 3.1 the Schema Object *is* a JSON Schema, where `examples` is an array
+and `example` is deprecated. When the `openapi` version is set to `3.1.x` the
+examples are therefore left untouched, and the same route generates:
+
+```json
+"/": {
+  "post": {
+    "requestBody": {
```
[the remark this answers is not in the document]
gevalo1: Fixed.

gevalo1:
Hi @Fdawgs

I see you triggered a copilot review, I've fixed the one comment that came out of that.
How can we progress this PR to have a chance of getting merged?

lib/spec/openapi/utils.js
```diff
@@ -222,7 +227,13 @@ const schemaTypeToNestedSchemas = {
   }
 }
 
-function resolveSchemaExamples (schema) {
+function resolveSchemaExamples (opts, schema) {
```
Eomm: usually `opts` is the last argument, I think we should switch them

lib/spec/openapi/utils.js
```diff
@@ -25,6 +25,10 @@ function prepareDefaultOptions (opts) {
   const hideUntagged = opts.hideUntagged
   const extensions = []
   const convertConstToEnum = opts.convertConstToEnum
+  // In OpenAPI 3.1 the Schema Object is a JSON Schema, where `examples` is an
+  // array of values and `example` is deprecated. Earlier versions require the
+  // OpenAPI specific `example` / `examples` fields instead.
```
Eomm: this explanation should live in the resolveSchemaExamplesRecursive's jsdoc

lib/spec/openapi/utils.js
gevalo1: Done.

lib/spec/openapi/utils.js
gevalo1: Moved.

PR #939, opened 2026-08-26
fix(openapi): convert type null and nullable array types to OpenAPI 3.0 nullable schemas

lx3133584:
## Problem

When using JSON Schemas containing `type: "null"` (e.g. generated by TypeBox `Type.Null()` or nullable type unions), OpenAPI 3.0 validation fails with `type must be equal to one of the allowed values` because `"null"` is not a primitive type keyword in OpenAPI 3.0.

## Root Cause

`convertJsonSchemaToOpenapi3` did not transform JSON Schema `type: "null"` or array types containing `"null"` into OpenAPI 3.0 `{ nullable: true }` syntax.

## Fix

- In `convertJsonSchemaToOpenapi3`, converted `type: "null"` to `{ nullable: true }`.
- Converted array types containing `"null"` (e.g. `type: ["string", "null"]`) into single type / `anyOf` with `nullable: true`.

## Testing

- Added unit tests in `test/spec/openapi/schema.test.js` validating schema conversion with `Swagger.validate()`.
- Verified 100% test coverage and TypeScript type tests pass.


Tony133:
Thanks for the PR. Closing for the following reasons:

- The conversion is applied regardless of the OpenAPI version. With `openapi: '3.1.0'`, `type: 'null'` and `type: ['string', 'null']` are valid and are currently emitted as-is; with this change they become `nullable: true`, which does not exist in 3.1, so the null information is lost. This is a regression for 3.1 users.
- The most common case, `anyOf: [{ type: 'string' }, { type: 'null' }]` (TypeBox `Type.Union([Type.String(), Type.Null()])`), becomes `anyOf: [{ type: 'string' }, { nullable: true }]`. In 3.0 `nullable` only has effect next to an explicit `type`, so the second branch matches anything and the whole schema accepts any value.
- For the same reason `{ type: 'null' }` becomes `{ nullable: true }` (unconstrained), and for multi-type unions `nullable` sits next to `anyOf` without a `type`, so it is ignored.
- The tests only assert the emitted shape, and `Swagger.validate()` checks document structure, not schema semantics.

PR #940, opened 2026-08-26
feat(openapi): support style and explode serialization for path parameters

lx3133584:
## Problem

Path parameters in OpenAPI 3.0 did not support `style` and `explode` serialization options, whereas query parameters already supported them.

## Root Cause

In `lib/spec/openapi/utils.js`, `plainJsonObjectToOpenapi3` forwarded `style` and `explode` properties on query/cookie/header containers, but omitted them for `case 'path'`.

## Fix

Forward `jsonSchema.style` and `jsonSchema.explode` to the OpenAPI path parameter definition when specified.

## Testing

- Added unit tests in `test/spec/openapi/schema.test.js` validating `style` and `explode` path parameter serialization with `Swagger.validate()`.
- Verified all 268 test suites (100% coverage) and TypeScript type tests pass cleanly.


Tony133:
Thanks for the PR. The path parameter change itself is correct and mirrors what is already done for query/header/cookie parameters. However, this branch also contains the unrelated `type: 'null'` → `nullable` conversion and its test from #939, which was closed because it regresses OpenAPI 3.1 output and produces unconstrained schemas in 3.0. It cannot be merged as is, so closing.

PR #941, opened 2026-08-27
feat(openapi): support name, id, summary, and description in example objects

lx3133584:
## Problem

When `examples` are specified on OpenAPI 3.0 schema definitions, `convertExamplesArrayToObject` unconditionally generated auto-incrementing keys (`example1`, `example2`) and wrapped the entire object under `{ value: ... }`. This made it impossible for users to specify custom example names/IDs or attach `summary`, `description`, or `externalValue` attributes as supported by the OpenAPI 3.0 Example Object specification.

## Solution

- Updated `convertExamplesArrayToObject` to inspect if an example element is an OpenAPI 3.0 Example Object (containing `value` or `externalValue`).
- When detected, preserves `summary`, `description`, and other Example Object properties, using `example.name` or `example.id` as the key (falling back to `exampleN`).
- Preserved backward compatibility for plain object and primitive examples.

## Testing

- Added unit tests in `test/spec/openapi/option.test.js` verifying preservation of summary, description, and custom example names/IDs.
- Verified all 267 unit tests pass with 100% statement, branch, function, and line coverage, and TypeScript tests pass cleanly.


Tony133:
Thanks for the PR. Closing because detecting Example Objects by the presence of `value`/`externalValue` changes the output for existing users whose payloads legitimately contain those keys, so this is a breaking change rather than an additive feature. It also doesn't cover the single-example path and allows `__proto__` as a key.

Note that named examples with `summary`/`description` are already supported through the `x-examples` field (see the [README](https://github.com/fastify/fastify-swagger#readme)), which is the explicit opt-in for this use case.

PR #943, opened 2026-08-28
fix(openapi): extract inline definitions and $defs into components.schemas

lx3133584:
## Problem

When using recursive schemas (such as TypeBox `Type.Recursive`) or schemas containing inline `definitions` or `$defs`, the generated OpenAPI 3.0 document rewrites internal `$ref` pointers (e.g. `#/definitions/def-0` to `#/components/schemas/def-0`), but deletes the underlying definition objects without placing them in `components.schemas`. This leaves dangling `$ref` references that fail OpenAPI document validation.

## Root Cause

`convertJsonSchemaToOpenapi3` previously unconditionally deleted `definitions` without preserving and registering them in `openapiObject.components.schemas`, and nested recursive conversion calls did not pass `openapiObject` down to child properties.

## Fix

- Updated `convertJsonSchemaToOpenapi3` to accept and thread `openapiObject` through recursive schema conversions.
- When `definitions` or `$defs` keys are encountered on a schema, their items are converted to OpenAPI 3.0 schemas and added to `openapiObject.components.schemas`.
- Supported `$defs` alongside `definitions` when normalizing `$ref` pointers to `#/components/schemas/`.

## Testing

- Added regression test in `test/spec/openapi/schema.test.js` validating that inline definitions are extracted to `components.schemas` and pass full Swagger/OpenAPI validation.
- Verified all 268 unit tests pass with 100% statement, branch, function, and line coverage.


Tony133:
Thanks for the PR. Closing for the following reasons:

- Extracted definitions are flattened into `components.schemas` with first-wins semantics. Shared schemas are registered there first and are named `def-0`, `def-1`, ... by default, so an inline `definitions['def-0']` is skipped and its `$ref` silently resolves to an unrelated shared schema. The same happens when two routes use the same local key. This produces a valid-looking document with the wrong schema, which is worse than the current dangling ref.
- `openapiObject` is not passed in the generic fallback recursion (`convertJsonSchemaToOpenapi3(opts, value)`), so definitions nested under `items`, `allOf`/`oneOf`, `additionalProperties`, etc. are still dropped.
- The branch also contains the unrelated example-object changes from #941, which was closed as a breaking change.

A proper fix needs namespacing or collision detection for the extracted keys.

PR #949, opened 2026-09-15
feat: OpenAPI 3.2.0 compatibility

### LINKED ISSUE #906: Add support for OpenAPI v3.2.0
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

OpenAPI Specification v3.2.0:
https://spec.openapis.org/oas/v3.2.0.html

Announcement:
https://www.openapis.org/blog/2025/09/23/announcing-openapi-v3-2

### Motivation

_No response_

### Example

_No response_

Tony133:
Proposal:

OpenAPI 3.2.0 is additive over 3.1 and uses the same JSON Schema dialect, so most of it already works: the version string is passed through and tag objects with `summary`, `parent` and `kind` are emitted unchanged. What was missing: the `$self` top-level property was dropped, the routes of custom HTTP methods could not be described, and the `QUERY` HTTP method (supported by Fastify) was working but untested.

Changes:

- passes `$self` through to the generated document
- with OpenAPI `3.2.0` or higher, the routes of the HTTP methods without a fixed field in the Path Item Object (e.g. `PROPFIND`, added with `fastify.addHttpMethod()`) are emitted under `additionalOperations`, keyed by the upper-case method as required by the specification. `query` is a fixed field in 3.2, so it stays where it is. With older versions the output is unchanged, since they cannot describe those methods at all
- adds an exported `OpenAPIV3_2` namespace built on top of the `openapi-types` 3.1 types: `Document` (`$self`), `TagObject` (`summary`, `parent`, `kind`), `ServerObject` (`name`) and `PathItemObject` (`query`, `additionalOperations`). It is accepted by the `openapi` option and it is both the input and the output type of `transformObject`
- adds unit tests for `$self`, hierarchical tags, a `QUERY` route with a request body and custom HTTP methods (3.2 and older versions), plus type tests
- documents in the README how to select the OpenAPI version, what is passed through and how custom HTTP methods are rendered

Note:

- `@apidevtools/swagger-parser` v13 rejects `openapi: 3.2.0` (it supports up to 3.1.2), so the new tests assert on the generated structure instead of calling `Swagger.validate`. That can be added once the parser supports 3.2.
- `$self` and the `query` field are passed through whatever the document version is: it is up to the user to set `openapi: '3.2.0'`.
- Other 3.2 additions that live in user-provided objects (`components.mediaTypes`, the OAuth2 device flow, `in: 'querystring'` parameters, `itemSchema`...) are passed through at runtime but are not typed yet: they can be added to the `OpenAPIV3_2` namespace incrementally, or dropped once `openapi-types` ships the 3.2 types.

Closes #906


mcollina (approved):
lgtm

lib/spec/openapi/utils.js
```diff
@@ -699,7 +701,31 @@ function prepareOpenapiSchemas (opts, jsonSchemas, ref) {
   return openapiSchemas
 }
 
+// Fixed fields of the Path Item Object holding an operation.
+// `query` has been added by OpenAPI 3.2.0
+const PATH_ITEM_METHODS = new Set(['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace', 'query'])
+
+// OpenAPI 3.2.0 added the `additionalOperations` map to the Path Item Object,
```
Fdawgs, line 708: @Tony133 missed this before it was merged. No issues with the code but this inline comment could have been a lot shorter and as a JSDoc tag, so it's visible when highlighting the function in an IDE.

Tony133: > @Tony133 missed this before it was merged. No issues with the code but this inline comment could have been a lot shorter and as a JSDoc tag, so it's visible when highlighting the function in an IDE.

No problem 😉👍🏻

The changes were a bit complex and I wanted to be specific; feel free to open a PR to reduce the comments if you wish. I'm trying to close open bugs since they've been sitting there for a long time.

Issue #861, opened 2025-03-20
Fields with Multiple Types


joshuamcginnis:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.2.1

### Plugin version

9.4.0

### Node.js version

22.9.0

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Unknown

### Description

This was previously reported in  #190 but was marked as stale and closed. I'd like to reopen it.

Basically, the issue is if you make a field that has multiple types, e.g.:

```ts
 "collectionName": {
  "type": ["null", "string"]
}
```

You get "Unknown Type: null,string"

![Image](https://github.com/user-attachments/assets/e4c4b1cd-36a5-4785-a90a-9287ebf7c4fe)

I think this has something to do with the underlying swagger implementation using OpenAPI 3.0+ which introduced `nullable: true` for defining nullable types. 

In my case, I'm actually using [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema) which outputs the schema (by default) in `jsonSchema7` format.

### Link to code that reproduces the bug

https://github.com/fastify/fastify-swagger/issues/190

### Expected Behavior

json schema types with multiple values, including nulls, should generate the correct swagger docs.

Tony133:
This is the same problem as #594 (and #190 before it): `type` arrays and `anyOf`/`oneOf` with a `null` branch are valid JSON Schema and valid OpenAPI 3.1, but the plugin emits them as-is even when the document targets OpenAPI 3.0.x, where they are invalid. 3.0 only knows `nullable: true`, which is why Swagger UI shows "Unknown Type: null,string".

**Workaround** in the meantime: target OpenAPI 3.1, where this syntax is valid and rendered correctly by recent Swagger UI versions:

```js
await fastify.register(require('@fastify/swagger'), {
  openapi: { 
     openapi: '3.1.0',
     /* ... */ 
  }
})
```

Closing as a duplicate; #594 stays as the tracking issue. The fix for both is to convert nullable unions into `nullable: true` when the target version is `< 3.1`.

Issue #190, opened 2019-06-17
Support for multiple types


L2jLiga:
When I tried to define type in schema like

```
{ type: ["string", "null"] }
```

Then I see in swagger:
```
"Unknown Type: string,null"
```

mcollina:
This should work, maybe you would like to send a PR to fix this?

L2jLiga:
I can try

Eomm:
This plugin works with swagger2 (OAS2) and the null type doesn't exist in [that version](https://swagger.io/specification/v2/). OAS3 has been introduced the `nullable` field for this purpose.

So I think you could archive this only with the static registration using an OAS3 file.
The `dynamic` registration (that I started to explore.. but no time to complete) need a refactor in order to generate an OAS3 descriptor because, right now, it produces an OAS2 configuration.

gwn:
I'm getting errors with `{ type: ["string", "number"] }` as well: `Unknown Type: string,number`

It looks like the value of the `type` key is stringified (note the comma between), so this may indicate that arrays aren't being handled. I may look more into this and submit a PR if I can find the time..

mcollina:
This module definitely needs some love!

Il giorno mer 31 lug 2019 alle 17:11 gwn <notifications@github.com> ha
scritto:

JacopoDaeli:
Yep. +1 on this as well. Will look into this ASAP.

vikasg603:
Hey, any update on this?
This still gives the error.

vikasg603:
Able to resolved this if we wanted to use null.
We can pass nullable: true to indicate schema can be nullable.

If anyone using https://github.com/ThomasAribart/json-schema-to-ts
I added a pr here: https://github.com/ThomasAribart/json-schema-to-ts/pull/37

valerioleo:
I can confirm that this error is still happening on 4.12.6
a JSON Schema containing

```
 "collectionName": {
          "type": [
            "null",
            "string"
          ]
        }
```

results in `"Unknown Type: null,string"`

PR #951, opened 2026-09-19
fix(openapi): support null types and type arrays in OpenAPI 3.0 documents

### LINKED ISSUE #594: Null type in JSON schema ignore when generating swagger.
### Prerequisites

### Fastify version

3.29.0

### Plugin version

6.0.1

### Node.js version

16.13.1

### Operating system

macOS

12.3.1

### Description

I'm using [Typebox](https://www.npmjs.com/package/@sinclair/typebox) to generate JSON schemas for an endpoint's response. In one endpoint some response properties may be `null`.

In Typebox we defined it as follow
```javascript
Type.Union([Type.String(), Type.Null()]
```
Which seems to generate the correct JSON schema
```json
{
  "anyOf": [
    { "type": "string" },
    { "type": "null" }
  ]
}
```

The generated swagger file does not indicate that field can be nullable (which cases some of our tests to fail).

### Steps to Reproduce

Create a schema with a property with type
```json
{
  "anyOf": [
    { "type": "string" },
    { "type": "null" }
  ]
}
```

Generate swagger file.

### Expected Behavior

_No response_

Tony133:
Proposal:

OpenAPI 3.0 has neither a `null` type nor the array form of `type`, so JSON Schemas using `type: 'null'`, `type: ['string', 'null']`, `type: ['string', 'number']` or an `anyOf`/`oneOf` with a `{ type: 'null' }` member produced invalid documents. This PR converts them to the 3.0 form, based on `nullable: true` and `anyOf`:

| JSON Schema | OpenAPI 3.0 |
|---|---|
| `{ anyOf: [{ type: 'string' }, { type: 'null' }] }` | `{ type: 'string', nullable: true }` |
| `{ type: ['string', 'null'] }` | `{ type: 'string', nullable: true }` |
| `{ type: 'null' }` | `{ type: 'object', nullable: true, enum: [null] }` |
| `{ type: ['string', 'number', 'null'] }` | `{ anyOf: [{ type: 'string' }, { type: 'number' }, { type: 'object', nullable: true, enum: [null] }] }` |
| `{ type: ['string', 'number'] }` | `{ anyOf: [{ type: 'string' }, { type: 'number' }] }` |
| `{ type: ['string'] }` | `{ type: 'string' }` |
| `{ anyOf: [{ $ref: 'Item#' }, { type: 'null' }] }` | `{ anyOf: [{ $ref: '...' }, { type: 'object', nullable: true, enum: [null] }] }` |

The conversion follows the OpenAPI 3.0.3 definition of `nullable`: a `true` value adds `null` to the allowed type only if `type` is explicitly defined within the same Schema Object. As a consequence:

- `null` alone is expressed as a nullable type restricted to `enum: [null]`, since a bare `{ nullable: true }` has no `type` to apply to (the type itself is irrelevant);
- a union is simplified to `{ type, nullable: true }` only when its single non-null member defines its own `type`. Otherwise (`$ref`, members without `type`, several members) the `null` member is kept in the union as a null schema;
- a `type` array is always converted, even when `null` is not one of the types, because the array form itself is not valid in 3.0. The null schema is added to the generated `anyOf` only when `null` was listed.

Sibling keywords are not lost: a member is hoisted into its parent only when it would not overwrite any keyword of the parent (e.g. `description`, `properties`), and a `type` array does not overwrite an existing `anyOf` (the generated union is added to `allOf` instead). Annotations on the null member (`title`, `description`, `x-*`) do not prevent its detection.

The conversion only applies to `3.0.x` documents, which is the default when no version is set. OpenAPI 3.1 uses JSON Schema 2020-12, where `type: 'null'` and `type` arrays are valid and `nullable` no longer exists, so 3.1 documents are left untouched (covered by a dedicated test). Swagger 2.0 mode is unchanged.

Note:
- Swagger 2.0 mode is unchanged: the Swagger 2.0 specification does not define nullable, anyOf/oneOf, the null type or the array form of type, so there is no valid form to convert these schemas to.
- Tests cover request bodies, responses, nested `items`, path/query/header parameters, shared schemas (both `addSchema` and `openapi.components.schemas`), keyword conflicts, and check that route schemas are not mutated. The documents generated by the conversion tests are validated with `swagger-parser`.
- The README has a new "Null types" section describing the conversion, including `type` arrays without `null`.
- This also covers the multiple-types case reported in #861 (and #190 before it), which was closed as a duplicate of #594.

Fixes #594


mcollina (approved):
lgtm

lib/spec/openapi/utils.js
```diff
@@ -507,6 +507,99 @@ function prepareOpenapiMethod (opts, schema, ref, openapiObject, url) {
   return openapiMethod
 }
 
+// OpenAPI 3.0 has no `null` type: nullability is expressed with `nullable: true`.
+// OpenAPI 3.1 uses JSON Schema 2020-12, where `type: 'null'` is valid and `nullable` no longer exists.
+// The document version defaults to 3.0.3 (see prepareOpenapiObject), so a
+// missing version must be treated as 3.0.
```
Fdawgs, line 513: @Tony133 missed this before it was merged. No issues with the code but this inline comment could have been a lot shorter and as a JSDoc tag, so it's visible when highlighting the function in an IDE.

lib/spec/openapi/utils.js
```diff
@@ -507,6 +507,99 @@ function prepareOpenapiMethod (opts, schema, ref, openapiObject, url) {
...
+function isAnnotationKeyword (key) {
+  return annotationKeywords.has(key) || key.startsWith('x-')
+}
+
+// OpenAPI 3.0.3: "A `true` value adds `"null"` to the allowed type specified by
+// the `type` keyword, only if `type` is explicitly defined within the same
+// Schema Object". A bare `{ nullable: true }` would therefore accept any value,
+// and `nullable: true` next to `anyOf`/`oneOf`/`$ref` would not accept `null`.
+// The only schema that accepts `null` and nothing else is a nullable type
+// restricted to `enum: [null]`; the type itself is irrelevant.
```
Fdawgs, line 533: Same as https://github.com/fastify/fastify-swagger/pull/951/changes#r4071016156

Tony133: The changes were a bit complex and I wanted to be specific; feel free to open a PR to reduce the comments if you wish. I'm trying to close open bugs since they've been sitting there for a long time.

lib/spec/openapi/utils.js
Tony133: see comment: https://github.com/fastify/fastify-swagger/pull/951#discussion_r4071056402

Issue #865, opened 2025-03-31
Recursive type definition using Typebox in schema leads to broken swagger file


CruelCrow:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.1.0

### Plugin version

9.4.2

### Node.js version

22.12.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

15.2

### Description

When using Recursive definition inside schema.body, swagger file generated as a result will be broken. It will generate `$ref: "#/components/schemas/def-0"`, but the "def-0" definition itself won't be generated.

The endpoint body is defined as follows:
```
import {Type, type Static} from '@sinclair/typebox';
const TreeNode = Type.Recursive(
    (Node) =>
        Type.Object({
            name: Type.String(),
            children: Type.Optional(Type.Array(Node)),
        }),
    {$id: 'Node'},
);

const BodySchema = Type.Object({
    name: Type.String(),
    age: Type.Number({minimum: 0}),
    treeNodes: Type.Array(TreeNode),
});
type BodyType = Static<typeof BodySchema>;

// ...

server.register(
    async (server: FastifyInstance) => {
        return [
            server.route({
                method: 'POST',
                url: 'submit',
                handler: async (request, reply) => {
                    const body = request.body as BodyType;
                    reply.send({message: `Received: ${body.name}, ${body.age}`});
                },
                schema: {
                    operationId: 'testSubmit',
                    summary: 'Test',
                    description: 'Submit user data',
                    body: BodySchema,
                    response: {
                        200: Type.Object({
                            message: Type.String(),
                        }),
                    },
                },
            }),
        ]
    },
    {prefix: '/'},
);
```

The definition of the "/submit" endpoint in this example will generated as such:
```
  /submit:
    post:
      operationId: testSubmit
      summary: Test
      description: Submit user data
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                name:
                  type: string
                age:
                  minimum: 0
                  type: number
                treeNodes:
                  type: array
                  items:
                    type: object
                    properties:
                      name:
                        type: string
                      children:
                        type: array
                        items:
                          $ref: "#/components/schemas/def-0"
                    required:
                      - name
                    title: Node
```
The problem is that the "def-0" schema definition will not be created in the resulting file.

### Link to code that reproduces the bug

[...]

PR #952, opened 2026-09-20
fix: support `$ref` to definitions of shared schemas

### LINKED ISSUE #639: JSON Schema Definitions not work

Tony133:
Proposal:

Swagger 2.0 and OpenAPI do not allow the `definitions` keyword inside a schema object, so the plugin drops it from shared schemas. The `$ref`s pointing into it were left untouched though, producing dangling references such as `#/definitions/def-0/definitions/foo`.

Moreover, the exact example of #639 (a nested `$id: '#address'`) currently throws `Cannot redefine property: Symbol(json-schema-resolver.refToDef)`, because the shared schemas are mapped again as `externalSchemas` on every `resolve()` call and their subschemas are evaluated against the wrong base URI.

This PR follows the approach suggested in the review of #676: instead of moving `definitions` somewhere else inside the schema, every nested definition is hoisted to the top-level `definitions` / `components.schemas`, and the references are rewritten through a map.

| Reference | Before | After |
|---|---|---|
| `http://foo/common.json#/definitions/foo` | `#/components/schemas/def-0/definitions/foo` (missing) | `#/components/schemas/def-0-foo` |
| `http://foo/common.json#address` | `#/components/schemas/def-0address` (missing) | `#/components/schemas/def-0-foo` |

In detail:

- nested `definitions` and `$defs` are hoisted as `<schema>-<key>`, at any depth (definitions nested in definitions, or in a property subschema); name collisions get a numeric suffix;
- all the `$ref`s of the final document are rewritten with a longest-prefix match, so `.../definitions/foo/properties/city` keeps its trailing pointer;
- local references of a shared schema (`#/definitions/foo`, `#/properties/bar`, `#`) are made absolute before the resolution, since `#` becomes the root of the whole document once the schema lands there;
- references to an anchor (`http://foo/common.json#address`, or `#address` from inside the shared schema) are converted to the JSON pointer of the anchored subschema before the resolution, since the ref resolver appends the fragment to the definition name as it is (`def-0address`). This is the form used by the Fastify "Fluent Schema" guide, mentioned in the comments of #639. The route schemas are cloned, not mutated, and only when at least one anchor exists;
- the fragment-only `$id`s are then removed from the cloned shared schemas: nothing points to them anymore, the ref resolver would list each of them as a duplicated definition (`def-1`), and Swagger 2.0 does not accept a nested `$id`;
- the schemas already known by the ref resolver are no longer passed as `externalSchemas`, which fixes the crash.

The walker is schema-aware: a property named `definitions`, or data inside `enum`/`default`/`examples`, is not mistaken for a keyword. The final rewrite walk only runs when something has been hoisted.

Known limitations:

- inline route schemas with their own `$id`/`definitions` are not covered (that is #865, it can build on this);
- an anchor is matched by the exact string `<$id of the shared schema>#anchor`: a relative reference that must be resolved against a different base URI is not recognized.

Note:

- Tests cover both Swagger 2.0 and OpenAPI modes with the exact schema of the issue, the shape of the Fluent Schema guide (references by anchor, by pointer and local anchors), anchors outside of `definitions`, nested definitions, `$defs`, local and recursive refs, name collisions and JSON pointer escaping. Generated documents are validated with `swagger-parser`.
- The README has a new section about the `definitions`/`$defs` of shared schemas, describing the `<schema>-<key>` naming and how to reference an anchor.

Fixes #639
Refs #675, #676


lib/util/definitions.js
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/**
+ * The ref resolver already knows the shared schemas: mapping them again as
+ * external schemas of a single `resolve()` call evaluates their subschemas
+ * against the wrong base URI, and it throws when one of those has a
+ * fragment-only `$id` (`{ $id: '#address' }`). Only the schemas set by the
+ * user in the plugin options must be provided.
+ */
+function unknownSchemas (schemas, ref) {
```
Fdawgs, line 151: Could be a lot shorter and as a JSDoc tag, so it's visible when highlighting the function in an IDE.

lib/util/definitions.js
```diff
@@ -0,0 +1,243 @@
...
+  }
+  return unknown
+}
+
+/**
+ * Swagger and OpenAPI do not allow the `definitions` and `$defs` keywords into
+ * a schema object. Every definition is moved out of the (already resolved and cloned)
+ * schema and returned, so it can be added to the top-level definitions.
+ * The `hoisted` map tracks `old pointer -> new name` to fix the references.
+ */
```
Fdawgs, line 168: Could be a lot shorter and as a JSDoc tag, so it's visible when highlighting the function in an IDE.

lib/util/definitions.js
Tony133: Done! see commit [4f58af8](https://github.com/fastify/fastify-swagger/pull/952/commits/4f58af8ca2d916b23a717a345263ece4b7c9d6cc)

lib/util/definitions.js
Tony133: Done! see commit [4f58af8](https://github.com/fastify/fastify-swagger/pull/952/commits/4f58af8ca2d916b23a717a345263ece4b7c9d6cc)

